PR #6, opened 2016-07-28
Resume chunked uploads

tobiasKaminsky:


src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
```diff
@@ -48,41 +51,49 @@
...
     private static final String TAG = ChunkedUploadRemoteFileOperation.class.getSimpleName();
-
-    public ChunkedUploadRemoteFileOperation(String storagePath, String remotePath, String mimeType){
-        super(storagePath, remotePath, mimeType);
-    }
+    private Context mContext;
 
     public ChunkedUploadRemoteFileOperation(
-            String storagePath, String remotePath, String mimeType, String requiredEtag
+            String storagePath, String remotePath, String mimeType, String requiredEtag, Context context
```
AndyScherzinger: As @przybylski mentioned in one of my PRs, as a general rule the context should always be the first parameter :)

src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
```diff
@@ -124,11 +135,28 @@ protected int uploadFile(OwnCloudClient client) throws IOException {
                         ", chunk index " + chunkIndex + ", count " + chunkCount +
                         ", HTTP result status " + status);
 
-                if (!isSuccess(status))
+                if (isSuccess(status)){
+                  successfulChunks.add(String.valueOf(chunkIndex));
+                } else {
+                    SharedPreferences.Editor editor = sharedPref.edit();
+                    editor.putStringSet(chunkId, successfulChunks);
+                    editor.commit();
```
AndyScherzinger: Not sure anymore, but I thought apply() is the preferred way since it works asynchronous.

AndyScherzinger:
Just some minor comments

tobiasKaminsky:
@AndyScherzinger please let us keep it with at least two thumbs up.
I know it is hard when we are only three devs, but I think it is worth the code quality, or?

AndyScherzinger:
@tobiasKaminsky yes, we should stay with the two thumbs up. Sorry for the impatience :(

PR #19, opened 2016-10-17
Increase the last chunk timeout in uploads

AndyScherzinger:
and also update to latest library versions.

cc: @tobiasKaminsky


AndyScherzinger:
@tobiasKaminsky this is also already shipped with the beta releases and can thus be merged since it has been tested/run in the wild

AndyScherzinger:
:+1:

PR #20, opened 2016-10-17
Set correct content-length for last chunk

iskradelta:
See https://github.com/nextcloud/android/issues/329

And possibly other reports/issues of "Unknown error"/retry-failing.


tobiasKaminsky:
@iskradelta this also applies to master branch, or?
If so, next time you can just open a PR against master and I will include it into beta for testing.

tobiasKaminsky:
We then, only in this case, need this PR also against master.

iskradelta:
Yes it does apply to master too. Aha, got it. I cherry picked them to master now, was that the right thing to do?

PR #22, opened 2016-10-19
Set correct content-length for last chunk (targeting master)

iskradelta:


AndyScherzinger:
Code looks good :+1: 
and has been test driven with positive feedback in the beta releases, thus will be merged after the build is done and green :)

AndyScherzinger:
@iskradelta merged! :rocket: Thank you very much for this contribution! It's now on the master and will be shipped with the next stable release of the app! :+1:

PR #23, opened 2016-10-27
Use channels for uploads and increase chunk size

iskradelta:
For https://github.com/nextcloud/android/issues/348

Faster uploads, yay.

Though Im still disappointed, but 4-5megabytes/s is good enough, pushed it from 800-900kbytes/s previously... but could have been just due to testing this during different days.

### DIFF
--- a/src/com/owncloud/android/lib/common/network/ChunkFromFileChannelRequestEntity.java
+++ b/src/com/owncloud/android/lib/common/network/ChunkFromFileChannelRequestEntity.java
@@ -29,15 +29,20 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
 import java.nio.channels.FileChannel;
+import java.nio.channels.WritableByteChannel;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
 
 import org.apache.commons.httpclient.methods.RequestEntity;
+import org.apache.commons.httpclient.util.TimeoutController;
 
 import com.owncloud.android.lib.common.utils.Log_OC;
+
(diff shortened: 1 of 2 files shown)


src/com/owncloud/android/lib/common/network/ChunkFromFileChannelRequestEntity.java
```diff
@@ -124,48 +129,42 @@ public void setmTransferred(long value) {
...
+                    if (size == 0) {
+                        size = -1;
                     }
+                    mDataTransferListener.onTransferProgress(transferredBytes, mTransferred, size, mFile.getAbsolutePath());
                 }
             }
 
-        } catch (IOException io) {
+        } catch (final IOException io) {
+            Log_OC.d(TAG, "Woopsie an io ", io);
```
AndyScherzinger: As for the logging, the if/elseif/else should have the respective log statements while the last else should have some generic message :wink:

AndyScherzinger:
Awesome @iskradelta - will the chunk size then be file-size aware since smaller files like photos or any other should still use smaller chunk size due to the fact that some users might operate on slow networks with data plans so this will impact their cost depending on the amount of failed/retried chunks.

@tobiasKaminsky please have a look since you are the owner of the beta :smiley:

mario:
@AndyScherzinger @tobiasKaminsky close this PR if it's no longer being worked on? :-/ Once we're out with 1.4.2, I'll write up what I want to do for 1.5 regarding this.

AndyScherzinger:
In general fine with me. @iskradelta fancy some collab with @mario on this topic? :)

iskradelta:
Whats the plan, what next?

AndyScherzinger:
> Whats the plan, what next?

Any thought @mario? I won't have the time in the near future to support this issue coding wise, plus you already thought about a solution here :)

PR #31, opened 2017-01-25
Change userinfo calls

mario:
This changes, updates and unifies getting the user information for both previous, current and future versions. Once this is merged, I have a PR for Android to update it to use the new calls.


AndyScherzinger:
@mario please see https://github.com/nextcloud/android/pull/502/files#diff-c197962302397baf3a4cc36463dce5eaR44 you can do the same thing here and open a PR for the app with the following in your gradle file:
```
compile 'com.github.nextcloud:android-library:change-userinfo-calls-SNAPSHOT'
```

Beware though that whenever you change your lib PR and pushed it to the remote you need to call:
```
https://jitpack.io/com/github/nextcloud/android-library/change-userinfo-calls-SNAPSHOT/build.log
```
In your browser which will trigger a new build of your lib PR and after that gradle will pick it up during building the app :)

As soon as this PR gets merged and a tag/release gets created we can trigger jitpack to build the release:
```
https://jitpack.io/com/github/nextcloud/android-library/1.0.11/build.log
```

mario:
@AndyScherzinger I am aware of that, and that's how I develop - the thing is that changes that use this in the Android branch are part of a user information feature, so I'll finish that and then push. Sounds good?

I'd appreciate a code review in the mean time.

AndyScherzinger:
@mario The code looks good to me, two questions/discussion though:

* do we need to add the parceler lib? It removes the need to write some boilerplate code, but enforces this dependency on app that use this library thus I would rather spend the time on writing the parcels, lib users will thank us for that I guess :)

* I this implementation then still backwards compatible so we can retrieve quota information  on older Nextcloud versions back to the original v9 and to make things even worse since the quota implementation in the app works this way for a reason also works for ownCloud 8 (which can't detect unlimited quota but that is not our problem)? cc @nickvergessen since that is how we started quota display last year :)

mario:
a) Unless you are very strongly against it, I would suggest we leave it. Implementing Parcelable is just a bunch of boilerplate and we can rather spend this time elsewhere. Also, Parceler is a well recognized and supported lib - not something I found in the hidden corners of internet.

b) Yes, this should work with both older and newer versions (yet to be released) with the changes we're implementing for the user information.

At the moment, if it's unlimited quota, we don't show that progress bar anyway.

AndyScherzinger:
> At the moment, if it's unlimited quota, we don't show that progress bar anyway.

Except for the case where we can't detect it. ownCloud doesn't inform us that a user has unlimited quota but sends us a quota = total disc space (so we show this as quota 😀)

> Unless you are very strongly against it, I would suggest we leave it.

I'd say we ask @tobiasKaminsky and @przybylski for their opinion. My point isn't about saving time but minimizing the amount of dependencies enforced on the library using app. Jackrabbit is needed to do the WebDAV part while parceler is a shortcut for us here which doesn't bring any feature/functionality to the lib user. Thus my point is rather from a architectural point of view.

mario:
Thank you for your comments, let's wait for them.

As far as I know, we're the only lib user, and while this doesn't bring anything for the end-user it does bring two things for us:

- more time for doing real things instead of writing boilerplate
- nicer code that is easier to maintain

If a library helps with that, I'm always up for using that.

AndyScherzinger:
> As far as I know, we're the only lib user

You are right actually. I thought that

* https://github.com/nerzhul/ownCloud-SMS-App/blob/master/build.gradle
* https://github.com/stefan-niedermann/nextcloud-notes/blob/master/build.gradle

would make use of the lib, but they actually don't 0_o

src/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.java
```diff
@@ -63,12 +107,27 @@ protected RemoteOperationResult run(OwnCloudClient client) {
         RemoteOperationResult result = null;
         int status = -1;
         GetMethod get = null;
+        String url;
+
+        OwnCloudBasicCredentials credentials = (OwnCloudBasicCredentials) client.getCredentials();
+
+        OwnCloudVersion version = client.getOwnCloudVersion();
+        boolean versionWithSelfAPI =
+                (version != null && version.isSelfSupported());
```
tobiasKaminsky: As we are using 120 as line wrap you can write this in one line, I guess.

test_client/src/com/owncloud/android/lib/test_project/TestActivity.java
```diff
@@ -338,11 +336,6 @@ public RemoteOperationResult removeShare(int idShare) {
 		
 	}
 
-    public RemoteOperationResult getQuota() {
-        RemoteGetUserQuotaOperation getUserQuotaOperation = new RemoteGetUserQuotaOperation();
-        return getUserQuotaOperation.execute(mClient);
-    }
-
```
tobiasKaminsky: Is it possible to write a real test case for this?

tobiasKaminsky:
I am really uncertain about using the library...
On one side it helps us as Mario says, but on the other hand it is in our core library, where it might be hard to remove if the library must be removed/changed (and we also have to maintain it).

test_client/src/com/owncloud/android/lib/test_project/TestActivity.java
mario: We could do a mockup test case because we don't have real servers with various versions and sample data deployed. But test project and sample client never really worked. Maybe sometime in the future I can work on automating server deployment and testing and all that stuff.

src/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.java
mario: Yea, dunno  how this happened. Fixed.

mario:
@tobiasKaminsky that's why I ALWAYS select a well maintained, recognized and used library - they're a good upstream, and it's easy to talk to them. In addition, there's really not much to maintain on our side, and it really cuts down on both the time and thinking of writing weird boilerplate.

Parceler, Butterknife, Dagger, Retrofit, and a few other libraries are in the arsenal of pretty much every decent developer, and somebody wrote them so we can benefit from the collective smarts :)

tobiasKaminsky:
As far as I understand the plan, it can happen that new user properties are added by the server admin.
How are they handled? Currently not, or?

mario:
Maybe I explained it wrong - if I did, sorry about that. What I meant is
that NC could add more fields in the future, not the admin :)

src/com/owncloud/android/lib/common/Quota.java
```diff
@@ -0,0 +1,87 @@
...
+ * Created by mdjanic on 24/01/2017.
+ */
+
+@Parcel
+public class Quota {
+    public long free;
+    public long used;
+    public long total;
+    public long quota;
+    public double relative;
```
mario: Yes. If they're private, then we incur a significant performance penalty due to reflection.

src/com/owncloud/android/lib/common/UserInfo.java
```diff
@@ -0,0 +1,127 @@
...
+public class UserInfo {
+    public String id;
+    public Boolean enabled;
+    public String displayName;
+    public String email;
+    public String phone;
+    public String address;
+    public String webpage;
+    public String twitter;
+    public Quota quota;
```
mario: Yes. If they're private, then we incur a significant performance penalty due to reflection.

AndyScherzinger:
Would new fields added by NC also mean a new server release each time?

mario:
@AndyScherzinger indeed.

mario:
Should be fine now @AndyScherzinger and @tobiasKaminsky.

AndyScherzinger:
👍

mario:
@AndyScherzinger @tobiasKaminsky merging this since you've both approved.

AndyScherzinger:
This change will now have to ship with 1.4.2 since #33 is needed for 1.4.2 and will also have to be merged. Is this an issue for the app?

mario:
@AndyScherzinger and you're right of course, I didn't think it through before merging.

We have two courses of action:

1) We merge user-info branch to the master for the app
2) I backport changes from that branch to master just to fix it to work with the new library

Let me know what you'd prefer!

AndyScherzinger:
I'd say we go with 2. since user-info is a v12 server feature afaik, so the app works like it does now but doesn't break with this lib change.

mario:
Well, it'll also work in a new 11.0.2, but yea ... gonna do it right away.

mario:
@AndyScherzinger https://github.com/nextcloud/android/pull/610

tobiasKaminsky:
I think the really right way would be to revert the merge.
And ship only bugfixes on library with 1.4.2.
And after releasing 1.4.2 merge this PR again.
(but we can stay with this. Just in case this happens again ;-) )

mario:
Yup, I would opt for staying with this now and make sure it does not happen
again :p

Go fix your bug :)

PR #33, opened 2017-01-30
fix bug preventing uploading the same large file twice

tobiasKaminsky:
The successful chunks were stored always, whereas they should only get stored if the upload is not successful.

Ref: https://github.com/nextcloud/android/issues/581

### DIFF
--- a/src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
+++ b/src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
@@ -161,11 +161,10 @@ public class ChunkedUploadRemoteFileOperation extends UploadRemoteFileOperation 
             }
 
         } finally {
+            SharedPreferences.Editor editor = sharedPref.edit();
             if (this.isSuccess(status)) {
-                SharedPreferences.Editor editor = sharedPref.edit();
                 editor.remove(chunkId).apply();
             } else {
-                SharedPreferences.Editor editor = sharedPref.edit();
                 editor.putStringSet(chunkId, successfulChunks).apply();
             }
 
--- a/src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
+++ b/src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
@@ -42,6 +42,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileChannel;
+import java.util.Calendar;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
@@ -70,7 +71,8 @@ public class ChunkedUploadRemoteFileOperation extends UploadRemoteFileOperation 
         RandomAccessFile raf = null;
 
         File file = new File(mLocalPath);
-        SharedPreferences sharedPref = mContext.getApplicationContext().getSharedPreferences("com.nextcloud.PREFERENCE_upload", Context.MODE_PRIVATE);
+        SharedPreferences sharedPref = mContext.getApplicationContext().
+                getSharedPreferences("com.nextcloud.PREFERENCE_upload", Context.MODE_PRIVATE);
         String chunkId = String.format("%08d", Math.abs(file.getName().hashCode()));
         Set<String> successfulChunks = sharedPref.getStringSet(chunkId, new LinkedHashSet<String>());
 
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
```diff
@@ -161,8 +161,13 @@ protected int uploadFile(OwnCloudClient client) throws IOException {
             }
 
         } finally {
-            SharedPreferences.Editor editor = sharedPref.edit();
-            editor.putStringSet(chunkId, successfulChunks).apply();
+            if (this.isSuccess(status)) {
+                SharedPreferences.Editor editor = sharedPref.edit();
+                editor.remove(chunkId).apply();
+            } else {
+                SharedPreferences.Editor editor = sharedPref.edit();
```
AndyScherzinger: Duplicate code, can be done before the if block

AndyScherzinger:
How are already saved uploads then handled? As in: Will this fix correct them too, does the user need to upload them again after the error and it works then or are they corrupted forever?

src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
```diff
@@ -161,8 +161,13 @@ protected int uploadFile(OwnCloudClient client) throws IOException {
             }
 
         } finally {
-            SharedPreferences.Editor editor = sharedPref.edit();
-            editor.putStringSet(chunkId, successfulChunks).apply();
+            if (this.isSuccess(status)) {
```
AndyScherzinger: wouldn't this part make line 158 obsolete since the finally block takes care of it anyways?

src/com/owncloud/android/lib/resources/files/ChunkedUploadRemoteFileOperation.java
```diff
@@ -162,7 +162,11 @@ protected int uploadFile(OwnCloudClient client) throws IOException {
 
         } finally {
             SharedPreferences.Editor editor = sharedPref.edit();
-            editor.putStringSet(chunkId, successfulChunks).apply();
+            if (this.isSuccess(status)) {
```
AndyScherzinger: the success part seems to be exactly the same as the last lines of the try {} part and the finally block will always be executed.

tobiasKaminsky:
> Will this fix correct them too, does the user need to upload them again after the error and it works then or are they corrupted forever?

No. 
But while thinking about it...The server will not store the uploaded chunks forever. I think this will be cleaned at least after 24h (@LukasReschke?)
So the best would be to add a date to the chunks. This would then also solve the problem for existing stored chunks.

tobiasKaminsky:
I hope I did not messed it up, I hast had 3 hours of sleep ;-)
I added a day string yyyy-mm-dd to identify the age of a chunk. If the chunk is not uploaded on the same day it will be ignored (as we have to assume that the chunk is deleted on server) and uploaded again.
If we upload the same chunk twice or even more to the server it should not be a problem.

This should also fix all older versions and allow to reupload the files as the day string is not included and therefore the chunks will be uploaded again.

All was tested on my emulator.

AndyScherzinger:
> If the chunk is not uploaded on the same day it will be ignored (as we have to assume that the chunk is deleted on server) and uploaded again.

What happens for Uploads which start at 23:58 and error on 00:05? You then will have chunks from "yesterday" and chunks from "today" of the same file and the same upload.

tobiasKaminsky:
Yes, in that case the resume on 00:06 or later will trigger a complete new upload.
The problem is that successfulChunks.contains(String.valueOf(chunkIndex + "_" + getDateAsString())) can only work on a defined key argument. I cannot find a chunk that was uploaded < 24 hours.
If this/your use case is something to consider the best would be to put the uploaded chunks in our sqlite database, where such an query is easily possible (or iterate over the array and test each chunk).

mario:
As long as we:

a) Don't end up with duplicate, hanging chunks I'm fine
b) As long as things work even if I try to upload things 5 times, I'm fine

And avoid complicating things further.

AndyScherzinger:
I agree with you guys! 👍 

So merge?

AndyScherzinger:
👍

mario:
LGTM - will merge, but will not make a new release. Will wait until you tell me I can merge the timestamp too.

mario:
@AndyScherzinger this repo still has LGTM?

AndyScherzinger:
Nope, but all PRs that got opened before we removed it still show it. No worries :)

PR #34, opened 2017-02-04
Added field for custom server name

Flole998:
Preparing everything to change title in the app as mentioned in nextcloud/android#240

### DIFF
--- a/src/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.java
+++ b/src/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.java
@@ -76,6 +76,7 @@ public class GetRemoteCapabilitiesOperation extends RemoteOperation {
     private static final String NODE_USER = "user";
     private static final String NODE_FEDERATION = "federation";
     private static final String NODE_FILES = "files";
+    private static final String NODE_THEMING = "theming";
 
     private static final String PROPERTY_STATUS = "status";
     private static final String PROPERTY_STATUSCODE = "statuscode";
@@ -102,6 +103,8 @@ public class GetRemoteCapabilitiesOperation extends RemoteOperation {
     private static final String PROPERTY_BIGFILECHUNKING = "bigfilechunking";
     private static final String PROPERTY_UNDELETE = "undelete";
     private static final String PROPERTY_VERSIONING = "versioning";
+    
+    private static final String PROPERTY_SERVERNAME = "name";
 
 
     /**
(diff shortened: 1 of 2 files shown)


src/com/owncloud/android/lib/resources/status/OCCapability.java
```diff
@@ -77,6 +78,7 @@ public OCCapability(){
         mVersionMicro = 0;
         mVersionString = "";
         mVersionString = "";
+        mServerName = "Nextcloud";
```
Flole998: This is the default value in the constructor. To avoid nullpointer exception we have to initialize it to something. This get's overwritten later if the field is found.

mario: I meant you can initialize with the app name, aka R.string.app_name :)

Flole998: Sorry I misunderstood you there. Of course that is possible.

AndyScherzinger: @mario this is library code, ```R.string.app_name``` is part of the app project, so we need to use a "hardcoded value" here.

mario: That is only half-true. In strings.xml (of the library) you create an app_name and put "Nextcloud" there. Once included in a project it will be overwritten with the app's strings.xml, and this is the approach we should take.

AndyScherzinger: Then you are right, so @Flole998 could you please add this string? :)

src/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.java
```diff
@@ -246,6 +249,14 @@ protected RemoteOperationResult run(OwnCloudClient client) {
                                     respFiles.getBoolean(PROPERTY_VERSIONING)));
                             Log_OC.d(TAG, "*** Added " + NODE_FILES);
                         }
+                        
+                        if (respCapabilities.has(NODE_THEMING)) {
+                            JSONObject respTheming = respCapabilities.getJSONObject(NODE_THEMING);
+                            // Add theming
+                            capability.setServerName(respTheming.getString(PROPERTY_SERVERNAME));
+                            Log_OC.d(TAG, "*** Added " + NODE_THEMING);
```
AndyScherzinger: Yes, see https://github.com/nextcloud/server/pull/1272 :)

src/com/owncloud/android/lib/resources/status/OCCapability.java
Flole998: It was not possible to do that as there is no context in that part of the library. I decided to leave the default value empty, that leaves handling up to the application using the library.

src/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.java
Flole998: Fields added

mario:
👍  from me. @AndyScherzinger ?

AndyScherzinger:
👍 also fine from my side (updated to latest master (merge due to GH click)

AndyScherzinger:
@mario I am fine with a merge, while due to the merge you would have to re-approve.

mario:
👍

mario:
Thanks for the contribution!

PR #37, opened 2017-02-10
Fix the build

mario:
This fixed the build.


tobiasKaminsky:
This can be closed as #38 will revert the too-early merge and thus this PR is not needed.

AndyScherzinger:
Thus delete the branch?

PR #25, opened 2016-11-22
Check for unshareable federated shares

AndyScherzinger:
cherry pick from oC for determine if a federated share is allowed to be re-shared

PR #44, opened 2017-03-01
Fix a missing method

mario:
This seems to be missing from master and we need it.


AndyScherzinger:
Erm, how about this one? https://github.com/nextcloud/android-library/pull/25 ? ;)

mario:
Right, deleting this one xD

PR #45, opened 2017-03-01
Fix LDAP login

mario:
This should fix LDAP login nd shouldn't affect others.

It should work on both 10 and pre 11.0.2 as well. Will test tomorrow.

I believe this is significant enough that it has to be in our next bugfix release.

### DIFF
--- a/src/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.java
+++ b/src/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.java
@@ -34,7 +34,6 @@ import com.owncloud.android.lib.common.UserInfo;
 import com.owncloud.android.lib.common.operations.RemoteOperation;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.utils.Log_OC;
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@mario can you explain a bit more how this fixes the LDAP login? Looking at the code you removed the self supported API check and now always go for the self support API and just check if there are values present. What would happen with a 9.0 instance (possibility of exceptions or just success==false)?

And yes it is also imho be significant enough to be included in the bugfix release.

mario:
@AndyScherzinger I will check all of this today by install 9 and 10.

The previous endpoint didn't really work well for LDAP powered logins because it looks like credentials.username was the wrong username! :D

mario:
@AndyScherzinger @tobiasKaminsky tested, all works as expected on 9 and 10.

AndyScherzinger:
👍 code looks good to me

mario:
@AndyScherzinger please re-approve.

@tobiasKaminsky go ahead and approve :P

AndyScherzinger:
:+1: re-approved. For future PR, can you rebase? :)

AndyScherzinger:
👍 code looks good to me

tobiasKaminsky:
👍 
as I do not have LDAP I cannot test it.

PR #46, opened 2017-03-02
Add support for new DAV endpoint

mario:
Adds support for a more modern DAV endpoint, needed for search and many other things in the future.

### DIFF
--- a/src/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
+++ b/src/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
@@ -24,6 +24,21 @@
 
 package com.owncloud.android.lib.common.operations;
 
+import android.accounts.Account;
+import android.accounts.AccountsException;
+
+import com.owncloud.android.lib.common.accounts.AccountUtils.AccountNotFoundException;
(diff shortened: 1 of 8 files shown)


AndyScherzinger:
👍 Looks code code wise 👍

mario:
@tobiasKaminsky any complaints?

mario:
@AndyScherzinger not working as-is, so closing. Needs rethinking a bit as the entire app doesn't have a central place to get webdav path.

AndyScherzinger:
@mario but we could update this then piece by piece, right? So this PR would still be legit and improve the implementation (over time), no?

mario:
@AndyScherzinger yes, but I think we need to sit and see what to do with this all before we start changing things :)

AndyScherzinger:
@mario this should be merged before #51 right? :) (else we stack PRs/branches again)

mario:
@AndyScherzinger no. I don't really want to merge this one as-is.

AndyScherzinger:
@mario so you want to merge it as part of #51?

mario:
@AndyScherzinger #51 does NOT include changes introduced with this PR, although it may look like it does :)

PR #50, opened 2017-03-09
Update Jackrabbit

mario:
For bugfix reasons. 

Also, after working with it for a bit now I really don't think Jackrabbit is our future.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ repositories {
 }
 
 dependencies {
-    compile 'org.apache.jackrabbit:jackrabbit-webdav:2.12.4'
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
👍 look good to me and yes the other part about a general solution which would then also cover ipv6 is a whole other topic :(

mario:
@AndyScherzinger removed reviewers? So you're ok with merging this as-is?

mario:
@tobiasKaminsky please approve.

AndyScherzinger:
Sorry for the late comment @mario and yes I am fine with merging as is

PR #53, opened 2017-03-24
Fix ciphers

mario:
Not sure if this is really used for connections (is it?), but if it is - this should fix ciphers issue.


AndyScherzinger:
Let's see if we can test drive this and afaik ```SSLSocketFactory ``` is in use for the connections :)

Now we need to find soemone who has a server for us that uses these stronger ciphers or at least check if it still works for our instances and push it to the beta branch of the lib and rebuild jitpack.

mario:
@AndyScherzinger I can set the cipher to whatever you want, as long as you put in the effort to find the ones that failed in our various issues :P

AndyScherzinger:
@mario that wouldn't help, because of the number of ciphers that might have been in use but never got reported, so for starter it seems worth a try for the beta channel to activate all available one, or not sure if we can get this info during handshake to then activate all ciphers supported by the respective server.

PR #55, opened 2017-03-27
Notifications

mario:
@AndyScherzinger this is now ready and working. Please merge.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath 'com.android.tools.build:gradle:2.3.0'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
     }
 }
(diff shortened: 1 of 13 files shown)


mario:
@AndyScherzinger this works. Ready for merging?

mario:
@AndyScherzinger do you know how to extend capabilities testing?

We could use testing for notifications?

        "notifications": {
          "ocs-endpoints": [
            "list",
            "get",
            "delete",
            "icons",
            "rich-strings"
          ]
        }

AndyScherzinger:
@mario actually no, I never did anything with capabilities, maybe Tobias did.

> We could use testing for notifications?

I didn't understand that one? at the moment we only use one endpoint, right? Even for v1 it might make sense to also have the delete operation? What do you think?

mario:
@AndyScherzinger we will never do delete as far as the endpoint is concerned (otherwise other clients won't get it) - once we start storing notifications in DB, we can do a "Remove notifications" and remove the from the DB however.

As for testing...dunno. It was late xD

AndyScherzinger:
👍 fine code wise and successfully tested

PR #60, opened 2017-05-15
Use app specific log folder

tobiasKaminsky:


src/com/owncloud/android/lib/common/utils/Log_OC.java
```diff
@@ -73,30 +74,26 @@ public static void wtf(String TAG, String message) {
...
-
+        try {
             // Create the current log file if does not exist
-            mLogFile.createNewFile();
+            isEnabled = mLogFile.createNewFile();
             mBuf = new BufferedWriter(new FileWriter(mLogFile, true));
-            isEnabled = true;
 
-            if (isFileCreated) {
+            if (isFolderCreated && isEnabled) {
```
AndyScherzinger: What if the folder already exists?

tobiasKaminsky: Should not happen as this is the app specific folder which will be deleted when the app is removed.
But I have added it, just to be safe.

AndyScherzinger:
👍 looks good to me

mario:
👍

PR #61, opened 2017-05-15
Activity RichSubject

aleister09:
This PR implement RichSubject to fix the issue 
[#971](https://github.com/nextcloud/android/issues/971)

### DIFF
--- a/src/com/owncloud/android/lib/resources/activities/models/RichSubject.java
+++ b/src/com/owncloud/android/lib/resources/activities/models/RichSubject.java
@@ -1,0 +1,58 @@
+/*  Nextcloud Android Library is available under MIT license
+ *   Copyright (C) 2017 Alejandro Bautista
+ *
+ *   @author Alejandro Bautista
+ *
+ *   Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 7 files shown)


mario:
Afaik rich objects are part of Activities v2 api which is related to the activities app. In case where we're still on v1 activities, it just won't have rich objects but I do agree we need to query capabilities to see what we have. 

As for OCS v2, it was (I think) available in NC9 so I think we're safe there.

src/com/owncloud/android/lib/resources/activities/models/RichSubject.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.ArrayList;
+
+/**
+ * RichSubject Data Model
+ */
+
+public class RichSubject {
+
+    public String richSubject;
+    ArrayList<RichObject> richObjectList=new ArrayList<>();
```
mario: Please make this List<RichObject> richObjecstList and some separation between name and = sign.

mario:
Commented a bit. Also additional comment:

Please rename RichSubject to RichElement or something because this also can be used for message_rich. Would appreciate if you add support for that as well.

AndyScherzinger:
👍 fine by me

PR #65, opened 2017-06-21
Bruteforce

tobiasKaminsky:
lib changes for: https://github.com/nextcloud/android/pull/1099

### DIFF
--- a/src/com/owncloud/android/lib/common/DynamicSessionManager.java
+++ b/src/com/owncloud/android/lib/common/DynamicSessionManager.java
@@ -5,7 +5,6 @@ import android.accounts.OperationCanceledException;
 import android.content.Context;
 
 import com.owncloud.android.lib.common.accounts.AccountUtils;
-import com.owncloud.android.lib.resources.status.OwnCloudVersion;
 
 import java.io.IOException;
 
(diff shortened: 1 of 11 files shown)


AndyScherzinger:
>  	remove version dependency as NC can this in all supported version 

would this drop some oC support? (as in oC server which aren't eol?)

tobiasKaminsky:
Prior oC 9.3. I am not completely sure what happens on older version, but hey we are NC ;-)

AndyScherzinger:
> Prior oC 9.3. I am not completely sure what happens on older version, but hey we are NC ;-)

true, but we also have an unknown number of oC users using the Nc client...So according to https://github.com/owncloud/core/wiki/Maintenance-and-Release-Schedule I think it would be good if we support oC 9.0 not sure if that would then work out of the box.

mario:
@tobiasKaminsky I think you should kill the commit which removes the version check. Simple to do, but with potential benefits. Then we can merge.

tobiasKaminsky:
You mean so that it can work (maybe) for oc?

mario:
Indeed. There's nothing else that this commit does or brings, except potentially removes compatibility with oC.

AndyScherzinger:
👍

mario:
👍

PR #67, opened 2017-07-06
Switch to ocsv2 for sharing to get the error messages directly

tobiasKaminsky:
https://docs.nextcloud.com/server/12/developer_manual/core/ocs-share-api.html

With v2 we get directly the error message and not first a ok/failure and then had to pick the error message.

### DIFF
--- a/src/com/owncloud/android/lib/resources/shares/CreateRemoteShareOperation.java
+++ b/src/com/owncloud/android/lib/resources/shares/CreateRemoteShareOperation.java
@@ -170,7 +170,7 @@ public class CreateRemoteShareOperation extends RemoteOperation {
     }
 
     private boolean isSuccess(int status) {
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
is v2 a 12+ feature? b/c then we would have to support v1 and v2 with server version check...

tobiasKaminsky:
According to Björn all NC version and even all oC version support v2.

AndyScherzinger:
Cool 🚀

mario:
Well, I think "all" means post 9.

tobiasKaminsky:
AIll supported, which is >=9 yes.

AndyScherzinger:
👍 next stop, approval by @mario :D

AndyScherzinger:
👍 next stop, approval by @mario :D

PR #68, opened 2017-07-10
Add missing transfer of favorite status

tobiasKaminsky:
While reading WebDav properties we get the status of isFavorite, but do not fransfer it to RemoteFile.


AndyScherzinger:
👍 nice catch!

mario:
👍

PR #69, opened 2017-07-23
Add lock failed error

mario:


AndyScherzinger:
@mario  So is lock failed a new status we can retrieve from the server?

mario:
@AndyScherzinger no, it's a local LOCK FAILED. We try to lock the file before uploading so it wouldn't get changed during the upload ;)

AndyScherzinger:
👍

tobiasKaminsky:
:+1:

PR #74, opened 2017-08-15
Public upload: permission needs to be 15, instead of 31.

tobiasKaminsky:
On newer nc version setting permission to 31 is not allowed as 16 is "share" permission.
So we have to use CRUD permission (15)


AndyScherzinger:
👍 fine by me

Issue #71, opened 2017-08-14
GPlay:   java.lang.NullPointerException - com.owncloud.android.lib.resources.status.OwnCloudVersion.<init>


AndyScherzinger:
java.lang.NullPointerException
com.owncloud.android.lib.resources.status.OwnCloudVersion.<init>

```
java.lang.NullPointerException: 
  at com.owncloud.android.lib.resources.status.OwnCloudVersion.<init> (OwnCloudVersion.java:84)
  at com.owncloud.android.lib.common.accounts.AccountUtils.getCredentialsForAccount (AccountUtils.java:210)
  at com.owncloud.android.lib.common.OwnCloudAccount.loadCredentials (OwnCloudAccount.java:126)
  at com.owncloud.android.lib.common.SimpleFactoryManager.getClientFor (SimpleFactoryManager.java:62)
  at com.owncloud.android.lib.common.DynamicSessionManager.getClientFor (DynamicSessionManager.java:42)
  at com.owncloud.android.lib.common.operations.RemoteOperation.run (RemoteOperation.java:282)
  at java.lang.Thread.run (Thread.java:764)
```

Hasn't been recorded before 2.0.0-rc2 but I guess this is due to a misconfigured server. Nevertheless I'd say we should handle it and gracefully fallback to the oldest server version we support. What do you think @mario @tobiasKaminsky? For whatever reason it seams the server version is null?!

tobiasKaminsky:
null version is a result of 
> am.getUserData(account, "oc_version") (am = AccountManager)

which returns 
> the user data, null if the account or key doesn't exist

So we should try to catch it there already, at least this is what we are doing on the other usages.

tobiasKaminsky:
Fix merged

PR #77, opened 2017-08-17
if no version info is available use oldest supported

tobiasKaminsky:
Ref: #71


AndyScherzinger:
👍 is 10 the oldest supported version? (I think so since 9 is end-of-life)

AndyScherzinger:
👍 re-approval

PR #80, opened 2017-08-31
Versioning app can be disabled

tobiasKaminsky:
which then leads to org.json.JSONException: No value for versioning.

If this Exception occurs the other remaining capabilites will not be set, which might lead to unexpected behaviour.

Edit: tested on a recent server and disabled all apps. Now there is no exception anylonger.


AndyScherzinger:
👍 nice catch!

@tobiasKaminsky do you think this should be put into the app's 2.0.0 release? Or postpone it to 2.0.1?

tobiasKaminsky:
2.0 as this might cause missing theming infos as they are fetched after the versioning info and thus might to one of the unknown crashes we had...

AndyScherzinger:
Alright! 👍  So waiting for @mario's feedback

mario:
Merge

PR #82, opened 2017-09-05
Fix timestamp

mario:


AndyScherzinger:
👍 works like a charm, necessary to fix the upload modification bug. cc @tobiasKaminsky

mario:
Awesome, thanks for testing this!

tobiasKaminsky:
What is the benefit of this? I miss the context :/

AndyScherzinger:
Files being uploaded set their modification date to the upload time...

mario:
And this fixed this! :D

AndyScherzinger:
this PR fixes that issue, so auto uploaded files will keep their modification date and also the server will only create an created activity entry right now auto upload leads to 2 activity entries: created (picture taken), modified (picture uploaded)

tobiasKaminsky:
:+1: 
Thanks for the explanatation :-)

tobiasKaminsky:
Do we need this for the "non-chunked" upload also?

AndyScherzinger:
> Do we need this for the "non-chunked" upload also?

Can you answer this @mario? I don't know...

mario:
@tobiasKaminsky that already works.

PR #83, opened 2017-09-05
Try to fix 0 byte uploads

mario:

On certain phones, .lenght() gives us 0 bytes. We then send anticipated size as 0 bytes, and when server receives more, it deletes the file as part of integrity check.

This should prevent the issue.


tobiasKaminsky:
@mario 
Shall this be in RC7?

mario:
No, does not work :(

PR #95, opened 2017-11-08
Add text-color and element-color to capabilities

tobiasKaminsky:
Title says it: adds text-color and element-color from theming api to capabilities.
If <NC13 this will set to "".


AndyScherzinger:
👍 looks good!

mario:
👍

PR #98, opened 2017-11-22
Fix transfer-encoding: chunked

tobiasKaminsky:
It seems to be much easier solveable than I thought. GetMethod already handles this correctly, but we had a check for the correct file length.

To check the correct file length we can use the file size in OCFile, but this then would have to be done on client/app side OR we need an additional propfind to get the size here.

I vote for keeping it simple at first here.
Later we can think about something file checksum based file checking.


AndyScherzinger:
👍 loooks good to me

mario:
👍

PR #99, opened 2017-11-22
Username with whitespace

tobiasKaminsky:
As pointed out by @jensd0e, nextcloud can correctly handle usernames with whitespaces in it.
But toggling favorite is not working as the whitespace has to be replaced with %20 as this results in a webdav url.

### DIFF
--- a/src/com/owncloud/android/lib/resources/files/ToggleFavoriteOperation.java
+++ b/src/com/owncloud/android/lib/resources/files/ToggleFavoriteOperation.java
@@ -68,21 +68,14 @@ public class ToggleFavoriteOperation extends RemoteOperation {
         }
 
         String webDavUrl = client.getNewWebdavUri(false).toString();
-        int pos = filePath.lastIndexOf('/') + 1;
-        filePath = filePath.substring(0, pos) + Uri.encode(filePath.substring(pos));
-
-        String fullFilePath = webDavUrl + "/files/" + userID + filePath;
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
👍 looks good!

mario:
Maybe do a full url encoding instead of the userId?

tobiasKaminsky:
> Maybe do a full url encoding instead of the userId?

There is a more complex logic currently written by you, which only encodes the file name, but not the folders. 
Is there a special reason you did it this way? I am unsure if folders with whitespaces are working currently?

https://github.com/nextcloud/android-library/blob/9a52aaf0c42658197d8f47ee5b9be8519866f989/src/com/owncloud/android/lib/resources/files/ToggleFavoriteOperation.java#L70-L72

Encoding the full path seems to be the correct solution for me?

mario:
I was referring to doing URL encoding of the user id, so please do that instead of a simple replace.

But you are right, we should encode the full URL too - please do it, and feel free to do it in this PR.

tobiasKaminsky:
Uri.encode would replace "/" by "%2F" and then the server endpoint does not understand the request.
e.g: user "admin", path "/delete/" --> /nc/remote.php/dav/files/admin%2Fdelete%2F

what we can do is to replace "%2F" with "/" after encoding...

mario:
Sure.

tobiasKaminsky:
@mario done and tested 👍

mario:
Merge

PR #101, opened 2017-12-06
Add NC13 version

tobiasKaminsky:
Needed to check if we use the newest server version.

After merging this, we can do a new lib release and then merge https://github.com/nextcloud/android/pull/1836


tobiasKaminsky:
Confirmed by mario at hackweek.

PR #102, opened 2017-12-21
Oreo: bump targets

tobiasKaminsky:


sample_client/build.gradle
```diff
@@ -9,7 +9,7 @@ dependencies {
 }
 
 android {
-    compileSdkVersion 25
+    compileSdkVersion 26
     buildToolsVersion '25.0.0'
```
AndyScherzinger: shoudl then always use 26.0.2

test_client/build.gradle
```diff
@@ -20,7 +20,7 @@ dependencies {
 }
 
 android {
-    compileSdkVersion 25
+    compileSdkVersion 26
     buildToolsVersion '25.0.0'
```
AndyScherzinger: same here, should always use 26.0.2

test_client/tests/build.gradle
```diff
@@ -6,7 +6,7 @@ dependencies {
 }
 
 android {
-    compileSdkVersion 23
+    compileSdkVersion 26
     buildToolsVersion "23.0.3"
```
AndyScherzinger: and here: 26.0.2

sample_client/build.gradle
tobiasKaminsky: https://developer.android.com/studio/releases/gradle-plugin.html

> As of gradle 3.0 plugin: With this update, you no longer need to specify a version for the build tools—the plugin uses the minimum required version by default. So, you can now remove the android.buildToolsVersion property.

Do we want to remove it? Less config options is always nice, or?

tobiasKaminsky:
Lint is complaining heavily for permissions, so I added them.
Those are the same as in our NC app.

AndroidManifest.xml
```diff
@@ -29,8 +29,13 @@
     android:versionName="1.0.13" >
 
     <uses-sdk
-        android:minSdkVersion="11"
-        android:targetSdkVersion="24" />
+        android:minSdkVersion="14"
+        android:targetSdkVersion="26"/>
+
+    <uses-permission android:name="android.permission.AUTHENTICATE_ACCOUNTS"/>
```
mario: Please put android:maxSdkVersion="22" for all of them.

tobiasKaminsky:
Ready for merge @mario

mario:
@tobiasKaminsky thanks. Please do the maxsdk dance in the app repo too :)

PR #105, opened 2018-01-09
Add support for IPv6

mario:
Fix IPv6 support! :)

Basically, if you have an AAAA record only it would work. But if you have A and AAAA it doesn't work.

This fixes it! :)

### DIFF
--- a/src/com/owncloud/android/lib/common/network/AdvancedSslSocketFactory.java
+++ b/src/com/owncloud/android/lib/common/network/AdvancedSslSocketFactory.java
@@ -178,7 +178,7 @@ public class AdvancedSslSocketFactory implements SecureProtocolSocketFactory {
         SocketAddress localaddr = new InetSocketAddress(localAddress, localPort);
 
         InetAddress address = InetAddress.getByName(host);
-        if (address instanceof Inet6Address) {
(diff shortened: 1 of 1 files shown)


mario:
cc @AndyScherzinger  ;)

tobiasKaminsky:
Do we have an ipv6 server to test?

mario:
No, but we have people willing to test? There are bunch of people reporting issues with IPv6?

tobiasKaminsky:
Test would be
- v4 only
- v6 only
- server which supports v4 and v6: use v4 client/app
- server which supports v4 and v6: use v6 client/app

If we do not have this, I'll set up something locally to test tomorrow.

tobiasKaminsky:
> No, but we have people willing to test?

Right, but I want to test it at least locally. As we do not have an independent dev version anymore, this would require a merge without testing...

tobiasKaminsky:
Testing right now

- [ ] server v4, client v4
- [ ] server v4, client v6
- [ ] server v6, client v4
- [ ] server v6, client v6
- [ ] server v4&v6, client v4
- [ ] server v4&v6, client v6

tobiasKaminsky:
Merging with only one approval as @AndyScherzinger is on vacation.

PR #106, opened 2018-01-11
Activity gson parsing: ignore key/value we do not know

tobiasKaminsky:
fix https://github.com/nextcloud/files_accesscontrol/issues/85

### DIFF
--- a/src/com/owncloud/android/lib/resources/activities/models/RichElementTypeAdapter.java
+++ b/src/com/owncloud/android/lib/resources/activities/models/RichElementTypeAdapter.java
@@ -90,26 +90,36 @@ public class RichElementTypeAdapter extends TypeAdapter<RichElement> {
         }
     }
 
-
-
-    RichObject readObject(String tag,JsonReader in) throws IOException {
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/lib/resources/activities/models/RichElementTypeAdapter.java
```diff
@@ -90,26 +90,36 @@ private void read(RichElement richElement, JsonReader in) throws IOException {
...
+                    richObject.setPath(in.nextString());
+                    break;
+                case "link":
+                    richObject.setLink(in.nextString());
+                    break;
+                case "server":
+                    richObject.setLink(in.nextString());
+                    break;
+                default:
+                    in.skipValue(); // ignore value
```
mario: This is, I believe, wrong. It will skip the next value. Instead, just don't do anything.

mario: Ignore me.

mario:
👍

PR #110, opened 2018-01-22
Fixes #109: Use correct field name when extracting user info fields

### LINKED ISSUE #109: Incorrect website json field fetched from server
in `ocs/v1.php/cloud/user` returned json has field named 'website' current implementation uses 'webpage' which always yields in not set website field.

przybylski:


mario:
👍

mario:
cc @AndyScherzinger

tobiasKaminsky:
NC11: webpage
NC12+: website

As NC11 is EOL I am fine with switching to website, but then please switch the naming on every occurrence, also on client, so we stay consistent.
We can put it into 3.1 then

mario:
@tobiasKaminsky if this is indeed true, I'd rather w support both even if 11 is technically EOL.

przybylski:
@tobiasKaminsky I wasn't aware of that. Thanks for pointing that out.

AndyScherzinger:
@tobiasKaminsky I'd second @mario and would also say it is better to still support Nc11 because even if it is not supported anymore the server would still work but now the client doesn't so supporting at least Nc13-11 seems a good compromise (where 10 is way to old for sure).

tobiasKaminsky:
Yeah, then I am fine with supporting NC11, but please name it website as this is the future :-)

AndyScherzinger:
👍

PR #112, opened 2018-01-25
We do not use quota right now, so do not fetch it

tobiasKaminsky:


AndyScherzinger:
Which quota is this?

tobiasKaminsky:
The quota for the specific file. If this is on an external storage it can happen that this quota is different from what we show in the drawer.
But as we do not use it, we can safely remove it. On sync of 5k files it reduces transferred size from 3.7 to 3.3Mb.

AndyScherzinger:
👍 sweet

mario:
I know this is against common sense, but I would rather see this code commented than removed :)

tobiasKaminsky:
For what? If we need it, we have it in git history. And as the code is really doing nothing (at least in our app ;-) ) it should be very safe to remove it.

mario:
👍

PR #113, opened 2018-01-27
Enhancement/use gson

przybylski:
* Reduce the complexity of hand parsing nodes, use gson
* Add a wrapper to avoid parsing server response

### DIFF
--- a/src/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.java
+++ b/src/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.java
@@ -38,6 +38,7 @@ import com.owncloud.android.lib.common.operations.RemoteOperation;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.ocs.OCSResponse;
+import com.owncloud.android.lib.ocs.ServerResponse;
 import com.owncloud.android.lib.resources.status.OwnCloudVersion;
 
 import org.apache.commons.httpclient.HttpStatus;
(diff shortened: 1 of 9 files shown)


AndyScherzinger:
:+1: Woohoo, nice one @przybylski

src/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.java
```diff
@@ -140,73 +124,25 @@ protected RemoteOperationResult run(OwnCloudClient client) {
...
-                JSONObject respOCS = respJSON.getJSONObject(NODE_OCS);
-                JSONObject respData = respOCS.getJSONObject(NODE_DATA);
+                Gson gson = new Gson();
+                JsonParser parser = new JsonParser();
+                JsonObject respJson = (JsonObject)parser.parse(response);
+                Type respType = new TypeToken<ServerResponse<UserInfo>>(){}.getType();
+                ServerResponse<UserInfo> ocsResponse = gson.fromJson(respJson, respType);
 
-                UserInfo userInfo = new UserInfo();
+                UserInfo userInfo = ocsResponse.ocs.data;
```
mario: Woot! What about using getters instead? getOcs().getData()? :)

przybylski: Both ways for me. not using setters mimics traversing JSON object

mario:
👍 

@przybylski ready for merging?

przybylski:
Yes and no. Its complete in terms of implementation. If we want to generalise it and use it on other classes then no.

mario:
As far as I'm concerned, we should use it all over the place.

przybylski:
I made few changes as a generalisation, have a look and let me know if we want to go that way.

mario:
@przybylski very good, yup. I went similar way with Talk (but used Retrofit & LoganSquare).

przybylski:
If we are ok with incrementally change all other operations (which I can do) than it is redy to review.

mario:
Yup, sure. 👍 
@AndyScherzinger can you ok one more time pls?

AndyScherzinger:
👍 sure :)

PR #115, opened 2018-02-06
Make OCSResponse a generic class, not just a class with geric method

przybylski:
Originally I falsly used a generic method, and assumed that it will work just like in C++. Java works a bit different when it comes to generics.

==
Sorry for introducing this bug https://github.com/nextcloud/android/issues/2101


AndyScherzinger:
👍 

cc @tobiasKaminsky @mario 

No worries @przybylski! Glad to have you back 🎉

mario:
👍

tobiasKaminsky:
Great, thank you for this quick fix :+1:

PR #118, opened 2018-02-11
Only log debug and verbose output for DEBUG builds.

### LINKED ISSUE #116: User information leaking through debug log prints.
adb logcat reveals several debug log entries marked with "GetRemoteUserInfoOperation". Here is one example.

`02-09 23:12:33.410 28516  3060 D GetRemoteUserInfoOperation: Successful response: {"ocs":{"meta":{"status":"ok","statuscode":100,"message":"OK","totalitems":"","itemsperpage":""},"data":{"enabled":"true","id":"xxx","quota":{"free":110406975803,"used":50654297797,"total":161061273600,"relative":31.45,"quota":161061273600},"email":"xxx@xxx.xx","phone":"","address":"","website":"https:\/\/redacted.com","twitter":"@redacted","groups":["admin"],"language":"en","display-name":"Redacted"}}}
`

It makes sense to print these as a debugging measure, but it doesnt belong in a production ready deployment. My recommendation would either be to use a wrapper class to filter the log prints away from RELEASE builds or use Proguard to filter out the debug code completely.

ardevd:
Ideally, verbose logging should not be compiled into an application except during development and debug logging should be stripped at runtime according to the Android developer documentation.

This PR at prevents debug and verbose output logging on non-DEBUG builds. 

Fixes #116

### DIFF
--- a/src/com/owncloud/android/lib/common/utils/Log_OC.java
+++ b/src/com/owncloud/android/lib/common/utils/Log_OC.java
@@ -3,6 +3,8 @@ package com.owncloud.android.lib.common.utils;
 import android.content.Context;
 import android.util.Log;
 
+import com.owncloud.android.lib.BuildConfig;
+
 import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileWriter;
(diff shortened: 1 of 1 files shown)


przybylski:
This and #116 can coexist. The bigger issue is that probably ProGuard is not correctly configured. It should take care of removing debug logs instead of doing it with if check.

ardevd:
I'll be happy to create a PR with ProGuard configuration that strips out the debug logs if that's a preferred solution for you.

ardevd:
Since you've already created a wrapper class for your logging function I think this PR is a good solution.

AndyScherzinger:
👍 I am fine with both PRs and would say that this is a "quick" fix while I agree that the real solution would be to go for ProGuard afterwards. Simply to get rid of any potentially expensive string concatenation within a debug statement which of course gets executed before the debug level gets checked.

cc @tobiasKaminsky @mario

tobiasKaminsky:
As we do not have to rush, I think the best way would be to go with proGuard and even revert #116.

mario:
What's the verdict here?

AndyScherzinger:
@mario @tobiasKaminsky what's your opinion?

I am fine with merging this PR as-is and let @ardevd go for the proGuard implementation also (in a separate PR, as stated).

ardevd:
I'll get started on the ProGuard stuff soon. However I'll have to make a PR for the Android project as well since it's there the actual obfuscation and stripping will take place.

tobiasKaminsky:
> debug logging should be stripped at runtime according to the Android developer documentation.

> (proguard)/It should take care of removing debug logs instead of doing it with if check.

I also think it is better to do it with ProGuard so that we do not even use the log calls.
I will close this, keep it as a reference and open up a new issue, so that we can tackle this at some point.

PR #119, opened 2018-02-12
Update ChunkedUploadRemoteFileOperation.java

mario:
IHMO, this is worth 3.0.1.

https://github.com/nextcloud/android/issues/2065


AndyScherzinger:
If ```"HTTP-OC-Chunked"``` needed the ```HTTP-``` being added couldn't the same apply for ```"OC-Chunk-Size"```? Just asking, since I don't have chunking configured in my actual setup atm. cc @tobiasKaminsky

mario:
@AndyScherzinger didn't test yet. I just implemented what our contributor suggested - but he now claims it doesn't work :( So we'll re-test this.

mario:
It seems like there are some changes between oC and Nc here.  I just updated the PR.

DagNygren:
It really doesn't work. Seems like the HTTP_ in front of the custom attributes is auto added by the server. Couldn't find any documents stating this but if you make the patch I suggested it will effectively disable the chunking on the server side (and that is why the transfer succeeded) and end up with chnuk-files instead of the original file. So don't do it! Sorry for the wrong diagnosis!

PR #125, opened 2018-02-16
If image size is zero, try to get size from media store

tobiasKaminsky:


mario:
What if we're uploading something that's not in the media store? :P 

In addition, really don't think this is the solution :-/

tobiasKaminsky:
This only happens only for one plus cameras, so far.
All of them are images or videos, so they should be in media store.

Citing myself from the PR in android:
> This is a fix for certain OnePlus devices, where the image size taken by the default camera is zero.
I tested it with many images, but still some are uploaded corrupted. This then due to a wrong size reported by media store (instead of 0, it is something between 20-100kb).
So it is better than nothing, but not 100% reliable. As this is only used if file.length() is not working, it should have no impact on other devices.

We can stall this PR until next Hackweek, when we can both try to find a better solution on my real device.

tobiasKaminsky:
closing as it is a fix for minor amount of users.

PR #128, opened 2018-02-20
UserId with whitespace: fixing search

tobiasKaminsky:
Otherwise this will fail:
https://github.com/nextcloud/android-library/blob/master/src/com/owncloud/android/lib/common/network/WebdavEntry.java#L74

as it would search for "tobi k" instead of "tobi%20k"


AndyScherzinger:
:+1:

mario:
👍

PR #132, opened 2018-02-21
Remote operation result might been null

tobiasKaminsky:
This happened to some users according to google play console.
Frankly, I do not get this, because we
- assign a value in try block
- if try block fails, we use catch and assign a value there

So this now returns a dummy result.
If you know a better way, please say :-)


AndyScherzinger:
:+1: no ideas about better ways :/

mario:
👍

PR #147, opened 2018-05-24
Gallery search

mario:
Videos + Photos need to be merged into one item as gallery. This adds library support for it.

cc @ardevd @AndyScherzinger @tobiasKaminsky

### DIFF
--- a/src/com/owncloud/android/lib/resources/files/SearchOperation.java
+++ b/src/com/owncloud/android/lib/resources/files/SearchOperation.java
@@ -213,7 +213,7 @@ public class SearchOperation extends RemoteOperation {
         Element folderElement = null;
         if (filterOutFiles) {
             folderElement = query.createElementNS(DAV_NAMESPACE, "d:is-collection");
-        }   
+        }
         Element equalsElement;
         if (searchType == SearchType.FAVORITE_SEARCH) {
             equalsElement = query.createElementNS(DAV_NAMESPACE, "d:eq");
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
code-wise it looks fine to me, is there a corresponding app branch where this is integrated?
On a general note @mario @tobiasKaminsky how do we handle this combined view for servers which simply don't create previews for video files? One way would be to just show the "default video type icon" while that doesn't look that nice but might be the only way...

mario:
@AndyScherzinger I just did a quick hack for the app to test this while developing, there's no corresponding app branch since someone else will have to do it.

The reason is that over a year ago we agreed to merge videos+photos into a Gallery both into iOS and Android, so I had to see how to do it. This PR is the result of it :)

AndyScherzinger:
@mario could you just push your hack to a branch on the app repo, that we I/we would have an impression of how to integrate and could pick it up from there :) That be of great help :)

tobiasKaminsky:
After release of 3.2 I can have a look to implement it in app.
But then we really need the "load only first 50/100 items" and then load next ones after reaching bottom, as with both pictures and videos the search will never ever finish…

mario:
Well, didn't we talk about search pagination and caching the proper way @icewind1991? :P

AndyScherzinger:
@tobiasKaminsky @mario for whatever reason the iOS client is super fast for this view?!

tobiasKaminsky:
There is already some work done by Roeland and me: https://github.com/nextcloud/android/pull/2250

AndyScherzinger:
> There is already some work done by Roeland and me

@tobiasKaminsky iOS is super fast now! Not needing any Nc14 stuff, so that doesn't matter it seems that this can be solved today...

tobiasKaminsky:
https://github.com/nextcloud/ios/issues/583 
I asked over there :-)

mario:
I need review on this @AndyScherzinger @tobiasKaminsky @ardevd

AndyScherzinger:
:+1:

tobiasKaminsky:
:+1:
I will have a look at the PR for app soon.

PR #148, opened 2018-05-30
Empty activites have 304 response

tobiasKaminsky:
Fixes https://github.com/nextcloud/android/issues/2658

 --> 

@AndyScherzinger @mario do we want to have this in 3.2? 

### DIFF
--- a/src/com/owncloud/android/lib/resources/activities/GetRemoteActivitiesOperation.java
+++ b/src/com/owncloud/android/lib/resources/activities/GetRemoteActivitiesOperation.java
@@ -134,7 +134,11 @@ public class GetRemoteActivitiesOperation extends RemoteOperation {
                 Log_OC.d(TAG, "Successful response: " + response);
                 result = new RemoteOperationResult(true, status, get.getResponseHeaders());
                 // Parse the response
-                activities = parseResult(response);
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1: I'd say let's put it in 3.2 since this might be misunderstood and lead to issue reports which we will then have to deal with so merging it for 3.2 seems to be the way with less effort :)

tobiasKaminsky:
@mario you agree? :-)

mario:
Yes, except I don't agree with the wording. "This stream"? What does that even means? :D

mario:
👍  otherwise.

AndyScherzinger:
The shown text is in the app's code so this PR is fine then :)

mario:
Sure. I was just saying change the text please, wherever :)

AndyScherzinger:
Server just says: No activity yet 
without a subtext

tobiasKaminsky:
At least on all our other "empty views" we have a subtext.

mario:
I like the fact that we have subtext, just not its actual content

AndyScherzinger:
> I like the fact that we have subtext, just not its actual content
I am also in favor of a subtext :+1:

PR #149, opened 2018-06-01
Add passwordProtected flag to OCShares

AndyScherzinger:
At the moment the server supports setting a password for a share link and for an email share. The fact that a share is actually password protected isn't reflected in the OCShare object.
For links that has been fine since if shared-with is set is is password protected  (even though this sounds strange...). This mechanism however won't work for email shares thus the client using the library should be able to parse/get/set this info and later store this fact to the database.

### DIFF
--- a/src/com/owncloud/android/lib/resources/shares/OCShare.java
+++ b/src/com/owncloud/android/lib/resources/shares/OCShare.java
@@ -99,6 +99,7 @@ public class OCShare implements Parcelable, Serializable {
     private long mUserId;
     private long mRemoteId;
     private String mShareLink;
+    private boolean mIsPasswordProtected;
     
     public OCShare() {
     	super();
@@ -133,6 +134,7 @@ public class OCShare implements Parcelable, Serializable {
         mUserId = -1;
         mRemoteId = -1;
         mShareLink = "";
+        mIsPasswordProtected = false;
     }	
     
     /// Getters and Setters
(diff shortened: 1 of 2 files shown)


src/com/owncloud/android/lib/resources/shares/OCShare.java
```diff
@@ -306,6 +316,7 @@ public void readFromParcel(Parcel source) {
         mUserId = source.readLong();
         mRemoteId = source.readLong();
         mShareLink = source.readString();
+        mIsPasswordProtected = source.readInt() == 0;
```
tobiasKaminsky: We are writing a 1 if mIsPasswortProtected is true in line 346, but compare with 0?
So if mIsPasswortProteced = true, then
on write: dest.writeInt(1).
on read: source.readInt() == 0 --> 1 == 0 --> false --> mIsPasswordProtected = false

tobiasKaminsky:
:-1: (this feels so harsh)

AndyScherzinger:
> :-1: (this feels so harsh)

All good :) Fixed your comment :) -> https://github.com/nextcloud/android-library/pull/149/commits/431cf90b611cd12e2c63e5823fc5472c8448dc57

AndyScherzinger:
@mario @tobiasKaminsky please review, this PR is needed by https://github.com/nextcloud/android/pull/2671

AndyScherzinger:
AWAITING 2ND PEER REVIEW SINCE 8 DAYS // AWAITING RE-REVIEW SINCE 5 DAYS

@mario @tobiasKaminsky please review, this PR is needed by https://github.com/nextcloud/android/pull/2671

@tobiasKaminsky well, please re-review since I fixed your comment :)

mario:
👍

tobiasKaminsky:
:+1:

PR #152, opened 2018-06-18
Fix double slash on search results with folder, e.g. "/test//folder//"

tobiasKaminsky:
Ref: https://github.com/nextcloud/server/issues/9892

Depending on the answer by server team, we still should add this as not all server get upgraded quick enough.


AndyScherzinger:
:+1:

PR #153, opened 2018-06-27
ignore unhandled rich objects

tobiasKaminsky:
Due to https://github.com/nextcloud/activity/issues/277, it can happen that we cannot parse an activity correctly. In these cases we try now to show as many infos as possible, setting the unhandled cases to "".
If there is now a problematic activity, it should show at least all other ones correctly. Prior to this, it showed no activity at all.


AndyScherzinger:
:+1:

PR #157, opened 2018-08-06
Add redirect info to external link

tobiasKaminsky:
Retrieves "redirect" info from api, default is "false".


AndyScherzinger:
pinging @mario for second review :)

AndyScherzinger:
Mario isn't available for reviews atm, thus merging with only one approval

PR #158, opened 2018-08-14
Return new eTag

tobiasKaminsky:
When checking for changed eTag, also return the new eTag, if it changed.


AndyScherzinger:
:+1:

PR #173, opened 2018-09-11
Skip pullapprove on dependencies labels

tobiasKaminsky:
Same as in android repo: if drone is fine, we merge the dependency so that it can be tested with next dev version.


AndyScherzinger:
:+1:

PR #179, opened 2018-09-24
Add note to OCShare

tobiasKaminsky:
Allows to show/use send on note, introduced on NC14.


AndyScherzinger:
:+1: still fine with me :)

PR #185, opened 2018-10-09
Log_OC: use class instead of TAG

tobiasKaminsky:
Previously we had to always create TAG, which is most of the time getClass().getSimpleName().

With this we can now use:
Log_OC.d(this, "log tag");
instead of
Log_OC.d(TAG, "log tag"); 

And we avoid creating TAG each and every time.

What do you think? If you are fine with this, I can add this for every log level. After that we can slowly get rid of TAG.


AndyScherzinger:
Hmmm, not sure about the performance impact :/

The static final Tag String is calculated when the vm loads the class and it is a single time cost plus accessing it via the reference while this PR would create the cost per logging call. 
The upside is that you now need to provide a class instead of a string which makes sure the log statement will always use a class name.

AndyScherzinger:
There is afaik also a logging framework for Android build by the guy who wrote the action bar sherlock and Butterknife. Just don't remember the name right now. Else we might even be able to do this via kotlin where we might not even have to have the class as a parameter for this to work.

Just some thoughts though.

tobiasKaminsky:
I guess it just depends how often log is called: if TAG is not used, it is wasted "time" to create the TAG.
If it called 100x it might be different…
(without knowing it, I am pretty sure/hope that VM is caching this somehow)

AndyScherzinger:
:+1:

tobiasKaminsky:
@AndyScherzinger I added this to the other logging methods and cleaned up the code

PR #189, opened 2018-10-25
Fix sample client

tobiasKaminsky:
I just found out that our sample client was not working, so I adjusted it to new project structure.


AndyScherzinger:
:+1:

PR #191, opened 2018-10-25
Revert api instead of compile

tobiasKaminsky:
As far as I understood from https://developer.android.com/studio/build/dependencies?utm_source=android-studio#dependency_configurations "compile" should have been replaced by "api", but this leads to failing client:

```
 error: package org.apache.commons.httpclient does not exist
error: cannot find symbol class GetMethod
```

So reverting this for now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Merging as this othewise kills all builds on client.

PR #209, opened 2018-12-18
add missing asset file

farindk:
The `sample_client` referenced a non-existing asset file. Without that file, the app appears non-functional.


tobiasKaminsky:
Thank you for your contribution.
Is this image license free? Otherwise I suggest that we use a Nexcloud logo.

farindk:
I just copied it over from the other example program from here: 'android-library/test_client/assets/imageFile.png'

PS: if you are worried about the licenses, you may also want to have a look at the video file here: 'android-library/test_client/assets/videoFile.MP4'. Looks like a movie excerpt.

tobiasKaminsky:
Successor is #209

PR #220, opened 2019-01-14
Bugfix: CreateFolderRemoteOperation sends wrong user-agent

farindk:
Bugfix: `CreateFolderRemoteOperation` sent the MKCOL request with the wrong user agent. Instead of the custom agent defined by `OwnCloudClientManagerFactory`, it always used the default nextcloud user-agent, as can be seen in the Apache log:

    admin [14/Jan/2019:15:37:55 +0100] "MKCOL /nextcloud/remote.php/webdav/ImageMeter/wurzel%205/test.imi HTTP/1.1" 201 1212 "-" "Jakarta Commons-HttpClient/3.1"
    admin [14/Jan/2019:15:37:55 +0100] "PROPFIND /nextcloud/remote.php/webdav/ImageMeter/wurzel%205/test.imi/thumb-640x512-Bild_vom_14._Jan._2019.jpg HTTP/1.1" 404 1525 "-" "ImageMeter/2.20.2"
    admin [14/Jan/2019:15:37:55 +0100] "PUT /nextcloud/remote.php/webdav/ImageMeter/wurzel%205/test.imi/thumb-640x512-Bild_vom_14._Jan._2019.jpg HTTP/1.1" 500 6002 "-" "ImageMeter/2.20.2"

### DIFF
--- a/src/com/owncloud/android/lib/resources/files/CreateFolderRemoteOperation.java
+++ b/src/com/owncloud/android/lib/resources/files/CreateFolderRemoteOperation.java
@@ -90,7 +90,6 @@ public class CreateFolderRemoteOperation extends RemoteOperation {
         MkColMethod mkCol = null;
         try {
             mkCol = new MkColMethod(client.getWebdavUri() + WebdavUtils.encodePath(mRemotePath));
-            client.setUseNextcloudUserAgent(true);
             client.executeMethod(mkCol, READ_TIMEOUT, CONNECTION_TIMEOUT);
 
             if (HttpStatus.SC_METHOD_NOT_ALLOWED == mkCol.getStatusCode()) {


tobiasKaminsky:
This is due to E2E which only works on Nextcloud.
So we force to use nc useragent on some operations.

As of 10.0.x we allow NC useragent on every server operation (server side),  so this should be not a problem.
Why do you need to change it?

tobiasKaminsky:
Then set your custom User-Agent on both:
https://github.com/nextcloud/android-library/blob/591eeb215c4cda44d479d1191202163dd225dd56/src/com/owncloud/android/lib/common/OwnCloudClientManagerFactory.java#L90-L92

https://github.com/nextcloud/android-library/blob/591eeb215c4cda44d479d1191202163dd225dd56/src/com/owncloud/android/lib/common/OwnCloudClientManagerFactory.java#L82-L84

This should do the trick, or?

AndyScherzinger:
@farindk Did the suggested solution mentioned by @tobiasKaminsky solve your issue?

farindk:
@AndyScherzinger Yes, for me, everything is fine, I was using a patched version anyways. But I still think it is a strange hack to use different user-agents for different actions. Other users will definitely fall into the same trap.

AndyScherzinger:
@tobiasKaminsky how to proceed with this PR?

tobiasKaminsky:
This PR will kill E2E.
At some point we can/will switch to NC user agent on entire app/library, so we can close this.

PR #225, opened 2019-01-17
Fix an issue with not populating all the fields

mario:
So I've been banging my head against the wall for hours wondering why don't I get all properties, and here it is!

I would have assumed this caused more bugs in the app, but no idea as we have so many places that we fill the RemoteFile in both the app and the library.

That should go into one place so things like this won't happen, but here is a fix for the current problem.

Please try to get me a library release out of this - thanks!

### DIFF
--- a/src/com/owncloud/android/lib/resources/files/model/RemoteFile.java
+++ b/src/com/owncloud/android/lib/resources/files/model/RemoteFile.java
@@ -232,10 +232,13 @@ public class RemoteFile implements Parcelable, Serializable {
         setRemoteId(we.remoteId());
         setSize(we.size());
         setFavorite(we.isFavorite());
+        setIsEncrypted(we.isEncrypted());
         setMountType(we.getMountType());
         setOwnerId(we.getOwnerId());
         setOwnerDisplayName(we.getOwnerDisplayName());
         setNote(we.getNote());
+        setUnreadCommentsCount(we.getUnreadCommentsCount());
+        setHasPreview(we.hasPreview());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/com/owncloud/android/lib/resources/files/model/RemoteFile.java
```diff
@@ -232,10 +232,13 @@ public RemoteFile(WebdavEntry we) {
         setRemoteId(we.remoteId());
         setSize(we.size());
         setFavorite(we.isFavorite());
+        setIsEncrypted(we.isEncrypted());
```
tobiasKaminsky: Only folders are marked as encrypted, thus this is only in ReadFolderRemoteOperation:
https://github.com/nextcloud/android-library/blob/56cdc0cf218af7104ace78f4c69e6072d2cd502f/src/com/owncloud/android/lib/resources/files/ReadFolderRemoteOperation.java#L175

So please remove this.

tobiasKaminsky:
> I would have assumed this caused more bugs in the app

Those are handled directly in the app, therefore no crashes.
But to be honest I also do not understand, why these functions are duplicated…

src/com/owncloud/android/lib/resources/files/model/RemoteFile.java
mario: Ergh, but a RemoteFile could be a folder too?

mario:
@tobiasKaminsky it's not about the crashes. It's about incorrect values being present.

src/com/owncloud/android/lib/resources/files/model/RemoteFile.java
tobiasKaminsky: Yes, but this is not used in FolderOperation and for files we do not need it as it is set by the parent folder.
So please remove this for now, and if we put all the different places together, we need to make sure that it will still work.

mario: Don't understand the reasoning I must admit? Like I told you, you can use ReadFileRemoteOperation to get folder information WITHOUT descending into all the children.

So this is indeed helpful to me.

tobiasKaminsky: Encryption API is not setting encryption flag on file base, but only on folders, as you cannot have an encrypted file without an encrypted parent folder. 
So this is just not needed. If you really want to have it in, please add an check if it is a folder.

PR #228, opened 2019-01-22
set note only if value is passed

tobiasKaminsky:
(e.g. on set note, which is only available in >=NC14)


AndyScherzinger:
@tobiasKaminsky also shows the licence acceptance issue :/ So drone image probably needs to be updated

tobiasKaminsky:
Yes, see here https://github.com/nextcloud/android-library/pull/229

PR #236, opened 2019-02-06
Cleanup activity

tobiasKaminsky:
remove non-public objects
remove setters, as Activity is "read-only" from server
added test case

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,11 +39,14 @@ configurations.all {
 }
 
 dependencies {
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
@tobiasKaminsky build fails, getter/setter could maybe done via lombok in case we already have lombok in the lib project (don't think so)

tobiasKaminsky:
Yeah, we can add lombok. After all it is only a build tool.
I'll do it tomorrow

tobiasKaminsky:
We should backport this to 3.5.0, at least the original problem: remove objects.

PR #243, opened 2019-02-11
Add support for delete a notification

tobiasKaminsky:

### DIFF
--- a/src/main/java/com/owncloud/android/lib/resources/notifications/DeleteNotificationRemoteOperation.java
+++ b/src/main/java/com/owncloud/android/lib/resources/notifications/DeleteNotificationRemoteOperation.java
@@ -1,0 +1,105 @@
+/* Nextcloud Android Library is available under MIT license
+ *
+ *   @author Tobias Kaminsky
+ *   Copyright (C) 2019 Tobias Kaminsky
+ *   Copyright (C) 2019 Nextcloud GmbH
+ *
+ *   Permission is hereby granted, free of charge, to any person obtaining a copy
+ *   of this software and associated documentation files (the "Software"), to deal
+ *   in the Software without restriction, including without limitation the rights
+ *   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *   copies of the Software, and to permit persons to whom the Software is
+ *   furnished to do so, subject to the following conditions:
+ *
+ *   The above copyright notice and this permission notice shall be included in
+ *   all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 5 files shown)


src/main/java/com/owncloud/android/lib/resources/notifications/DeleteNotificationRemoteOperation.java
```diff
@@ -0,0 +1,105 @@
...
+                Log_OC.e(TAG, "Failed response while getting user notifications ");
+                if (response != null) {
+                    Log_OC.e(TAG, "*** status code: " + status + " ; response message: " + response);
+                } else {
+                    Log_OC.e(TAG, "*** status code: " + status);
+                }
+            }
+        } catch (Exception e) {
+            result = new RemoteOperationResult(e);
+            Log_OC.e(TAG, "Exception while getting remote notifications", e);
```
AndyScherzinger: Didn't we change that so you can provide the class object instead of a string?

AndyScherzinger:
Findbugs score increased :/

tobiasKaminsky:
> HTTP_PARAMETER_POLLUTION: HTTP Parameter Pollution
Concatenating unvalidated user input into a URL can allow an attacker to override the value of a request parameter. Attacker may be able to override existing parameter values, inject a new parameter or exploit variables out of a direct reach. HTTP Parameter Pollution (HPP) attacks consist of injecting encoded query string delimiters into other existing parameters. If a web application does not properly sanitize the user input, a malicious user may compromise the logic of the application to perform either client-side or server-side attacks.
In the following example the programmer has not considered the possibility that an attacker could provide a lang such as en&user_id=1, which would enable him to change the user_id at will.
>
>Vulnerable Code:
String lang = request.getParameter("lang");
GetMethod get = new GetMethod("http://www.host.com");
get.setQueryString("lang=" + lang + "&user_id=" + user_id);
get.execute();
Solution:
Sanitize user input before using it in HTTP parameters.

This is meant:
https://github.com/nextcloud/android-library/blob/12e38054440bef25403b6b0e44fabb84603ed924/src/main/java/com/owncloud/android/lib/resources/notifications/GetNotificationRemoteOperation.java#L79-L88

But url is somewhat hardcoded. Yes id can be anything, but only an integer.
Also this is checked on server side if this is valid/user is authorized.

AndyScherzinger:
👍 fine with me, just waiting for the build to turn green

AndyScherzinger:
@tobiasKaminsky build failes:
`` 
FAILURE: Build failed with an exception.

* What went wrong:
Task 'jacocoTestDebugUnitTestReport' not found in root project 'android-library'.
```

PR #256, opened 2019-03-05
provide more infos on toString()

tobiasKaminsky:
Is something missing?

### DIFF
--- a/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
+++ b/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
@@ -61,6 +61,8 @@ import java.util.Locale;
 
 import javax.net.ssl.SSLException;
 
+import lombok.ToString;
+
 
 /**
  * The result of a remote operation required to an ownCloud server.
@@ -70,6 +72,7 @@ import javax.net.ssl.SSLException;
  *
  * @author David A. Velasco
  */
+@ToString
 public class RemoteOperationResult implements Serializable {
 
     // Generated - should be refreshed every time the class changes!!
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
I'd say that looks fine to me, what do you think @ezaquarii ?

ezaquarii:
Looks ok.

For the future reference, if you are not concerned about aesthetics, you can make your life a bit easier with a `toString` builder. I believe IDE can autogenerate it for you too using a standard template.

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/ToStringBuilder.html

`toString()`, `equals()` and `hashCode()` are tedious to maintain... :(

AndyScherzinger:
Well, regarding code generation, depending on what should be part of the output we could also use https://projectlombok.org/features/ToString since we already use lombok.

tobiasKaminsky:
Thank you both for your suggestions.
Lombok seems to be very cool as the toString options are bound to code, so one does not have to remember to re-generate anything, but Lombok is doing this.

AndyScherzinger:
BEware though that as per my understanding lombok will only consider fields for the toString code. If that is enough then lombok would solve it :)

tobiasKaminsky:
> lombok will only consider fields for the toString code

Therefore I added @ToString.Include to getLogMessage() so that this will also be included.

tobiasKaminsky:
```
junit.framework.AssertionFailedError: RemoteOperationResult(mSuccess=false, mHttpCode=401, mHttpPhrase=Unauthorized, mException=null, mCode=UNAUTHORIZED, getLogMessage=Operation finished with HTTP status code 401 (fail))
at junit.framework.Assert.fail(Assert.java:50)
at junit.framework.Assert.assertTrue(Assert.java:20)
```

PR #269, opened 2019-03-22
show sharees

tobiasKaminsky:
use userId for uid_owner


tobiasKaminsky:
Spotbugs increased due to this:

> CC_CYCLOMATIC_COMPLEXITY: Method is excessively complex
This method has a high cyclomatic complexity figure, which represents the number of branch points. It is likely difficult to test, and is brittle to change. Consider refactoring this method into several to reduce the risk.

As we cannot change this in short time, I'll increase FindBugs value.

PR #277, opened 2019-04-15
Enhance userId

tobiasKaminsky:
- clarify loginName, displayName, userId
- migrate userId to OwncloudClient


src/androidTest/java/com/owncloud/android/lib/common/OwnCloudClientManagerTest.java
```diff
@@ -0,0 +1,50 @@
+package com.owncloud.android.lib.common;
```
AndyScherzinger: Should test classes also have a MIT license header?

PR #323, opened 2019-07-28
Plug custom logger implementation into Log_OC

ezaquarii:
This changes makes Log_OC implementation swappable at runtime, so app using this library can plug their own logger.

This is required to fix https://github.com/nextcloud/android/issues/4228


ezaquarii:
I need to reopen this.

PR #324, opened 2019-07-28
Plug custom logger implementation into Log_OC

ezaquarii:
Needs https://github.com/nextcloud/android/pull/4275

This changes makes Log_OC implementation swappable at runtime,
so app using this library can plug their own logger.

### DIFF
--- a/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
+++ b/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
@@ -14,6 +14,82 @@ import java.util.Locale;
 
 
 public class Log_OC {
+
+    private static Adapter impl = new LegacyImpl();
+
+    /**
+     * This adapter allows external library users to hook up
+     * application-specific logging framework, redirecting calls to legacy
+     * logging functions to new or custom implementation.
+     */
+    public interface Adapter {
+        void i(String tag, String message);
+        void d(String tag, String message);
+        void d(String tag, String message, Exception e);
+        void e(String tag, String message);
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
```diff
@@ -35,72 +111,73 @@ public static String getLogPath() {
...
+    }
+
     public static void i(String TAG, String message) {
-        Log.i(TAG, message);
-        appendLog(TAG + ": " + message);
+        impl.i(TAG, message);
     }
 
     public static void i(Object object, String message) {
         i(object.getClass().getSimpleName(), message);
```
tobiasKaminsky: I added this some time ago to not have to use TAG over and over again.
Do we want to have this also in the new interface?

ezaquarii: Yeah, I used to do that too - it simplifies on-desk debugging, but doesn't work well for field deployments.

My change is meant to be backward compatible and since this is a library and I cannot know who uses it, I didn't dare to change the interface.

Ideally, when diagnosing issues, we'd like ot have more sophisticated tagging by subsystem, such as "sync" or "mediaplayer", to narrow down logs to a particular domain, not only a single class. This requires a bit different approach.

I definitely plan to revisit this, but I'm not ready to comment.

tobiasKaminsky: Ok, then we keep it this way.

PR #327, opened 2019-08-15
Let appendPhoneInfo use new system

tobiasKaminsky:
Previously it used old "appendLog" which created old log file.


AndyScherzinger:
Same here, 2nd re-review & approval from @ezaquarii needed ❤️

PR #328, opened 2019-08-19
do not use ocClient for anonymous connections

tobiasKaminsky:

### DIFF
--- a/src/main/java/com/owncloud/android/lib/resources/notifications/RegisterAccountDeviceForNotificationsOperation.java
+++ b/src/main/java/com/owncloud/android/lib/resources/notifications/RegisterAccountDeviceForNotificationsOperation.java
@@ -75,8 +75,8 @@ public class RegisterAccountDeviceForNotificationsOperation extends RemoteOperat
 
     @Override
     protected RemoteOperationResult run(OwnCloudClient client) {
-        RemoteOperationResult result = null;
-        int status = -1;
+        RemoteOperationResult result;
+        int status;
         PushResponse pushResponse;
         PostMethod post = null;
 
@@ -109,11 +109,9 @@ public class RegisterAccountDeviceForNotificationsOperation extends RemoteOperat
                     result = new RemoteOperationResult(false, status, post.getResponseHeaders());
                 }
             }
-
         } catch (Exception e) {
             result = new RemoteOperationResult(e);
             Log_OC.e(TAG, "Exception while registering device for notifications", e);
-
         } finally {
             if (post != null) {
                 post.releaseConnection();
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/lib/resources/notifications/RegisterAccountDeviceForNotificationsOperation.java
```diff
@@ -75,8 +75,8 @@ public RegisterAccountDeviceForNotificationsOperation(String pushTokenHash, Stri
 
     @Override
     protected RemoteOperationResult run(OwnCloudClient client) {
-        RemoteOperationResult result = null;
-        int status = -1;
+        RemoteOperationResult result;
+        int status;
```
AndyScherzinger: due to this change the initial value is `0` while it has been `-1` before, is this on purpose?

src/main/java/com/owncloud/android/lib/resources/notifications/RegisterAccountDeviceForProxyOperation.java
```diff
@@ -62,10 +64,9 @@ public RegisterAccountDeviceForProxyOperation(String proxyUrl, String pushToken,
         this.userPublicKey = userPublicKey;
     }
 
-    @Override
-    protected RemoteOperationResult run(OwnCloudClient client) {
-        RemoteOperationResult result = null;
-        int status = -1;
+    public RemoteOperationResult run() {
+        RemoteOperationResult result;
+        int status;
```
AndyScherzinger: same here, now `0` instead of `-1`

src/main/java/com/owncloud/android/lib/resources/notifications/UnregisterAccountDeviceForProxyOperation.java
```diff
@@ -59,10 +61,9 @@ public UnregisterAccountDeviceForProxyOperation(String proxyUrl, String deviceId
         this.userPublicKey = userPublicKey;
     }
 
-    @Override
-    protected RemoteOperationResult run(OwnCloudClient client) {
-        RemoteOperationResult result = null;
-        int status = -1;
+    public RemoteOperationResult run() {
+        RemoteOperationResult result;
+        int status;
```
AndyScherzinger: same here, now `0` instead of `-1`

src/main/java/com/owncloud/android/lib/resources/notifications/RegisterAccountDeviceForNotificationsOperation.java
tobiasKaminsky: Hm. AndroidStudio said that this is overwritten in any case:
status is written in L95 and read in L98, so it should be safe

src/main/java/com/owncloud/android/lib/resources/notifications/RegisterAccountDeviceForProxyOperation.java
tobiasKaminsky: It is always set: 
https://github.com/nextcloud/android-library/blob/9b552a4947cd88f7fdb90fa8ccdebcab6b084dda/src/main/java/com/owncloud/android/lib/resources/notifications/RegisterAccountDeviceForProxyOperation.java#L83-L86

(except if there is an error before, but then status is never read)

PR #338, opened 2019-09-23
Defensive check for Log_OC arguments for null

ezaquarii:
Log.d does not handle null, causing NPE thrown from
native logger implementation. This defensive check
ensures that legacy clients don't crash the system.

### DIFF
--- a/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
+++ b/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
@@ -42,51 +42,69 @@ public class Log_OC {
     static class LegacyImpl implements Adapter {
 
         @Override
-        public void i(String tag, String message) {
-            Log.i(tag, message);
-            appendLog(tag + ": " + message);
+        public void i(final String tag, final String message) {
+            if (tag != null && message != null) {
+                Log.i(tag, message);
+                appendLog(tag + ": " + message);
+            }
         }
 
         @Override
-        public void d(String tag, String message) {
-            Log.d(tag, message);
-            appendLog(tag + ": " + message);
+        public void d(final String tag, final String message) {
+            if (tag != null && message != null) {
+                Log.d(tag, message);
+                appendLog(tag + ": " + message);
+            }
         }
 
         @Override
-        public void d(String tag, String message, Exception e) {
-            Log.d(tag, message, e);
-            appendLog(tag + ": " + message + " Exception: " + Arrays.toString(e.getStackTrace()));
+        public void d(final String tag, final String message, Exception e) {
+            if (tag != null && message != null && e != null) {
+                Log.d(tag, message, e);
+                StackTraceElement[] trace = e.getStackTrace();
+                appendLog(tag + ": " + message + " Exception: " + Arrays.toString(trace));
+            }
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


ezaquarii:
This fixes all issues similar to this one once and for all:
https://github.com/nextcloud/android/issues/4542

ezaquarii:
This PR does **not** need a client change.

src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
```diff
@@ -42,51 +42,69 @@
...
 
         @Override
-        public void d(String tag, String message, Exception e) {
-            Log.d(tag, message, e);
-            appendLog(tag + ": " + message + " Exception: " + Arrays.toString(e.getStackTrace()));
+        public void d(final String tag, final String message, Exception e) {
+            if (tag != null && message != null && e != null) {
+                Log.d(tag, message, e);
+                StackTraceElement[] trace = e.getStackTrace();
+                appendLog(tag + ": " + message + " Exception: " + Arrays.toString(trace));
```
tobiasKaminsky: Is there a reason you separated this? Could also still be inline, or?

src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
```diff
@@ -42,51 +42,69 @@
...
 
         @Override
-        public void e(String tag, String message, Throwable t) {
-            Log.e(tag, message, t);
-            appendLog(tag + ": " + message + " Exception: " + Arrays.toString(t.getStackTrace()));
+        public void e(final String tag, final String message, final Throwable t) {
+            if (tag != null && message != null && t != null) {
+                Log.e(tag, message, t);
+                StackTraceElement[] trace = t.getStackTrace();
+                appendLog(tag + ": " + message + " Exception: " + Arrays.toString(trace));
```
tobiasKaminsky: Same here

tobiasKaminsky:
@ezaquarii thanks again for your PR and finding the cause of the problem :tada: 
I have just one question and then we can merge it.

ezaquarii:
Can you pls elaborate what do you mean by "separation"?

BTW this implementation is invalid. I shoud use logical or operator, not and.

src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
ezaquarii: I'm not sure what you mean my "separation" vs "inline"?

ezaquarii: The logger implementation in `Log_OC` was extracted so we can use new logger implementation without re-writing modifying world. The application swaps this implementation for it's own `Logger`, but the legacy code (other apps?) can still use the old implementation.

https://github.com/nextcloud/android/pull/4275

tobiasKaminsky: Previously it was:
Arrays.toString(e.getStackTrace()))

now it is:
StackTraceElement[] trace = e.getStackTrace();
Arrays.toString(trace)

ezaquarii:
Line became too long and linter complained. :)

PR #346, opened 2019-10-16
Handle empty/null/invalid activity response better

tobiasKaminsky:

### DIFF
--- a/src/main/java/com/owncloud/android/lib/resources/activities/GetActivitiesRemoteOperation.java
+++ b/src/main/java/com/owncloud/android/lib/resources/activities/GetActivitiesRemoteOperation.java
@@ -32,6 +32,7 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.google.gson.JsonSyntaxException;
 import com.google.gson.reflect.TypeToken;
 import com.owncloud.android.lib.common.OwnCloudClient;
 import com.owncloud.android.lib.common.operations.RemoteOperation;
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/lib/resources/activities/GetActivitiesRemoteOperation.java
```diff
@@ -180,17 +181,27 @@ public boolean hasMoreActivities() {
...
+                    .registerTypeAdapter(PreviewObject.class, new PreviewObjectAdapter())
+                    .create();
+            Type listType = new TypeToken<List<Activity>>() {
+            }.getType();
 
-        return gson.fromJson(jsonDataArray, listType);
+            return gson.fromJson(jsonDataArray, listType);
+
+        } catch (JsonSyntaxException e) {
+            return new ArrayList<>();
```
AndyScherzinger: Should at least have a log statement so we can see what happened instead of just geeting an empty result (like in the first line...)

tobiasKaminsky: Good idea.
I tried it a moment ago, but then the test fails, as Log_OC is not mocked.
I have a look.

tobiasKaminsky: Done

AndyScherzinger:
@tobiasKaminsky quota test keeps failing...

tobiasKaminsky:
> @tobiasKaminsky quota test keeps failing...

Should work now. Branch was too old, so I did a rebase.

PR #350, opened 2019-10-29
Add content search type

tobiasKaminsky:
Successor of #319 


tobiasKaminsky:
Question is what to do with that? ;-)

AndyScherzinger:
> Question is what to do with that? ;-)

@tobiasKaminsky what would we be using it for?

tobiasKaminsky:
> @tobiasKaminsky what would we be using it for?

We can search then by mimeype, e.g. get all pdf or all videos.
I updated https://github.com/nextcloud/android/pull/4748, so better talk there as it is in the client code.

PR #373, opened 2019-12-12
Add Parcelable interface to OwnCloudAccount and OwnCloudVersion

ezaquarii:
Parcelable interface is required to allow migration
of legacy code using raw platform Account passed
via Intent. To migrate those code paths, new User
model must be passed via Intent extras as well.

Since OwnCloudAccount and *Version are used (temporarily)
as implementation details of User, we need to make those
POJOs Parcelable as well.


src/main/java/com/owncloud/android/lib/common/OwnCloudAccount.java
```diff
@@ -135,4 +137,72 @@ public String getDisplayName() {
             return null;
         }
     }
+
+    /*
+     * Autogenerated equals and hash
```
AndyScherzinger: Could also be generated via lombok

src/main/java/com/owncloud/android/lib/common/OwnCloudBasicCredentials.java
```diff
@@ -71,4 +73,63 @@ public boolean authTokenExpires() {
     public String toOkHttpCredentials() {
         return Credentials.basic(username, authToken);
     }
+
+    /*
+     * Autogenerated equals and hash
```
AndyScherzinger: Lombok, see above

AndyScherzinger:
Looks good to me, equals and hashcode could be auto generated at compile time though.

src/main/java/com/owncloud/android/lib/common/OwnCloudBasicCredentials.java
ezaquarii: @AndyScherzinger Thanks, I remved explicit implementation and enabled Lombok generator.

src/main/java/com/owncloud/android/lib/common/OwnCloudAccount.java
```diff
@@ -135,4 +139,42 @@ public String getDisplayName() {
...
+     */
+
+    @Override
+    public int describeContents() {
+        return 0;
+    }
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags) {
+        dest.writeParcelable(this.baseUri, flags);
```
tobiasKaminsky: is "this" needed?
We have this in no other place in our code and it makes it harder to read.

src/main/java/com/owncloud/android/lib/common/OwnCloudAccount.java
```diff
@@ -135,4 +139,42 @@ public String getDisplayName() {
...
+        dest.writeParcelable(this.credentials, flags);
+        dest.writeString(this.displayName);
+        dest.writeString(this.name);
+        dest.writeParcelable(this.savedAccount, flags);
+    }
+
+    protected OwnCloudAccount(Parcel in) {
+        this.baseUri = in.readParcelable(Uri.class.getClassLoader());
```
tobiasKaminsky: Same here

src/main/java/com/owncloud/android/lib/common/OwnCloudAccount.java
ezaquarii: @tobiasKaminsky 

It is not neede per se, but it is autogenerated code and that is the kind of template Android Studio plugin uses.

I removed it - not everybody will use this particular plugin so we can assume it will be modified manually.

src/main/java/com/owncloud/android/lib/common/OwnCloudAccount.java
ezaquarii: removed `this.` as well

ezaquarii:
> Looks good to me, for some reason (can't check on the phone) spotbugs seems unhappy pensive

@tobiasKaminsky 
@AndyScherzinger 

I compared those 2 reports.

1. 2 dodgy code warnings caused by Lombok-generated hashCode (return value stored in local variable); we shall consider excluding those from SpotBugs as we can't control code generator here
2. 3 correcness warnings about uninitalzied arrays returned from `Parcelable` implementation; since `Parcelable` implementation uses 2 stage initalizaton (array allocation then value is set), this is a false positive

I don't have any sensible recommendation on how to deal with this - those checks have some value, on the other hand there is really nothing particularly wrong with this code.

ezaquarii:
I seems that l still need to add some more parcelables to legacy POJOs...

tobiasKaminsky:
I am fine with this :+1: 
@ezaquarii just increase manually /scripts/analysis/findbugs-results.txt to 168 and then we can merge it.

ezaquarii:
I increased spotbugs threshold to 168

ezaquarii:
@tobiasKaminsky This change is backward compatible - no client change is required.

tobiasKaminsky:
Hm. This checks hangs. Anyway, I merge it now.

PR #380, opened 2020-01-11
Make response member in OkHttpMethodBase nullable

ezaquarii:
The class uses non-null lateinit variable that
is not initialized. The member should be of
nullable type.

Fixes nextcloud/android#5073
Fixes nextcloud/android#5102
Fixes nextcloud/android#5100
Fixes nextcloud/android#5142
Fixes nextcloud/android#5156

### DIFF
--- a/src/main/java/com/nextcloud/common/NextcloudClient.kt
+++ b/src/main/java/com/nextcloud/common/NextcloudClient.kt
@@ -107,9 +107,9 @@ class NextcloudClient(var baseUri: Uri, val context: Context) : OkHttpClient() {
                     destination = redirectedDestination
 
                     if (getRequestHeader("Destination").isNullOrEmpty()) {
-                        method.requestHeaders.put("destination", destination)
+                        method.addRequestHeader("destination", destination)
                     } else {
-                        method.requestHeaders.put("Destination", destination)
+                        method.addRequestHeader("Destination", destination)
                     }
                 }
                 status = method.execute(this)
--- a/src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
+++ b/src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
@@ -39,10 +39,14 @@ import okhttp3.Response
  */
 abstract class OkHttpMethodBase(var uri: String,
                                 val useOcsApiRequestHeader: Boolean) {
-    lateinit var response: Response
-    var queryMap: Map<String, String> = HashMap()
-    var requestHeaders: MutableMap<String, String> = HashMap()
-    var requestBuilder: Request.Builder = Request.Builder()
+    companion object {
+        const val UNKOWN_STATUS_CODE: Int = -1;
+    }
+    
+    private var response: Response? = null
+    private var queryMap: Map<String, String> = HashMap()
+    private val requestHeaders: MutableMap<String, String> = HashMap()
+    private val requestBuilder: Request.Builder = Request.Builder()
 
     fun OkHttpMethodBase() {
         requestHeaders.put("http.protocol.single-cookie-header", "true")
@@ -56,32 +60,53 @@ abstract class OkHttpMethodBase(var uri: String,
         return httpBuilder.build()
     }
 
+    /**
+     * Set request headers completely replacing existing headers.
+     * To clear request headers, call this method with empty list.
+     *
+     * @param headers List of header-value pairs
+     */
+    fun setRequestHeaders(vararg headers: Pair<String, String>) {
+        requestHeaders.clear()
+        requestHeaders.putAll(headers)
+    }
+
+    /**
+     * Adds request header, overwriting any existing value.
+     *
+     * @param header HTTP request header name
+     * @param value HTTP request header value
+     */
+    fun addRequestHeader(header: String, value: String) {
+        requestHeaders.put(header, value)
+    }
+
     fun setQueryString(params: Map<String, String>) {
         queryMap = params
     }
 
     fun getResponseBodyAsString(): String {
-        return response.body()?.string() ?: ""
+        return response?.body()?.string() ?: ""
     }
 
     fun releaseConnection() {
-        response.body()?.close()
+        response?.body()?.close()
     }
 
     fun getStatusCode(): Int {
-        return response.code()
+        return response?.code() ?: UNKOWN_STATUS_CODE
     }
 
     fun getStatusText(): String {
-        return response.message()
+        return response?.message() ?: ""
     }
 
     fun getResponseHeaders(): Headers {
-        return response.headers()
+        return response?.headers() ?: Headers.Builder().build()
     }
 
     fun getResponseHeader(name: String): String? {
-        return response.header(name)
+        return response?.header(name)
     }
 
     fun execute(nextcloudClient: NextcloudClient): Int {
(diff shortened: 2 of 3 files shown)


src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
```diff
@@ -39,7 +39,7 @@ import okhttp3.Response
  */
 abstract class OkHttpMethodBase(var uri: String,
                                 val useOcsApiRequestHeader: Boolean) {
-    lateinit var response: Response
+    private var response: Response? = null
     var queryMap: Map<String, String> = HashMap()
```
ezaquarii: @tobiasKaminsky Is there a reason why those members are public?

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
```diff
@@ -39,7 +39,7 @@ import okhttp3.Response
...
     var requestHeaders: MutableMap<String, String> = HashMap()
     var requestBuilder: Request.Builder = Request.Builder()
```
ezaquarii: those could be `val`

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
```diff
@@ -56,32 +56,53 @@ abstract class OkHttpMethodBase(var uri: String,
         return httpBuilder.build()
     }
 
+    /**
+     * Set request headers completely replacing existing headers.
+     * To clear request headers, call this methot with empty list.
```
AndyScherzinger:
```suggestion
     * To clear request headers, call this method with empty list.
```

src/main/java/com/nextcloud/common/NextcloudClient.kt
```diff
@@ -107,9 +107,9 @@ class NextcloudClient(var baseUri: Uri, val context: Context) : OkHttpClient() {
                     destination = redirectedDestination
 
                     if (getRequestHeader("Destination").isNullOrEmpty()) {
-                        method.requestHeaders.put("destination", destination)
+                        method.setRequestHeaders("destination" to destination)
```
AndyScherzinger: This changes the behavior since the original code just replaces the destination value, while this call leads to a `requestHeaders.clear()` clearing the whole header collection.

AndyScherzinger: So `addRequestHeader(...)` should lead to the same behavior afaik.

tobiasKaminsky: Indeed, it should be added, but not remove all other headers.

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
```diff
@@ -56,32 +56,53 @@ abstract class OkHttpMethodBase(var uri: String,
...
     }
 
     fun releaseConnection() {
-        response.body()?.close()
+        response?.body()?.close()
     }
 
     fun getStatusCode(): Int {
-        return response.code()
+        return response?.code() ?: 0
```
tobiasKaminsky: 0 as status code for failing response seems to be wrong, or?
Maybe 400?

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
```diff
@@ -103,7 +124,7 @@ abstract class OkHttpMethodBase(var uri: String,
         if (nextcloudClient.followRedirects) {
             return nextcloudClient.followRedirection(this).getLastStatus()
         } else {
-            return response.code()
+            return response?.code() ?: 0
```
tobiasKaminsky: same here? 400 instead of 0?

src/main/java/com/owncloud/android/lib/resources/activities/GetActivitiesRemoteOperation.java
```diff
@@ -130,8 +130,7 @@ public RemoteOperationResult run(NextcloudClient client) {
 
             status = client.execute(get);
             String response = get.getResponseBodyAsString();
-
-            String nextPageHeader = get.response.header("X-Activity-Last-Given");
+            String nextPageHeader = get.getResponseHeader("X-Activity-Last-Given");
```
tobiasKaminsky: Nice enhancement :+1:

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
tobiasKaminsky: No, I just forgot that the default visibility in Kotlin is different than in Java.

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
tobiasKaminsky: ok :-)

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
tobiasKaminsky: I wanted to avoid null, and thus I used lateinit, but seems to be wrong?

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
AndyScherzinger: 400 seems to be the best guess while I can't say if is is always a client error and if bad request is always right, but as a fallback, I'd say it is fine.

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
ezaquarii: Yes, it is incorrect use. Keyword latetinit is a workaround for situations where you want to declare a non-nullable value, but it cannot be initialized immediately. Kotlin requires all variables to be initailized in a primary constructor (this eliminates whole class of bugs). When this is not possible (for example, when you can set it in `onCreate()`, you can annotate a variable `lateinit`. Kotlin will generate some code that checks if variable is initialized before first use and throw NPE if not, but the variable is still non-nullable.

This is not an alternative for `java.util.Optional` or nullable type.

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
ezaquarii: The whole problem whith the API is that it tries to get some data from non-existing entity.
There is no response object so there is no status code as well and there is also no good answer what to do with that.
0 seems to be a good choice for a 'non-value'. Apache is using -1. Good as well.

400 is not universally valid proposition - what is resposne is null because theere were not network? In such case, no http request was ever made, so there is no code, headers, body, etc.
Whatever we do, 0 or -1 is still better than NPE, IMO.

tobiasKaminsky: Yeah, 400 would indicate that at least something was tried, but it is more like an uninitialized state, and thus -1 makes more sense.

tobiasKaminsky:
Ready from my side :+1

AndyScherzinger:
@tobiasKaminsky feel free to merge (from my pov)

ezaquarii:
@tobiasKaminsky I squashed the branch, anticipating some tests updates, but I see no unit tests.
If this is all we shall do here, I'm done as well.

PR #394, opened 2020-02-03
Fix exception handling in NextcloudClient

### LINKED ISSUE #393: NextcloudClient ignores exceptions
It looks like exception handling in `NextcloudClient.execute()` is not implemented at all. This causes the client to fail due to unhandled exceptions due to random, genuine and expected network errors.

Related issues:

nextcloud/android#5245
nextcloud/android#5305
nextcloud/android#5310
nextcloud/android#5331
nextcloud/android#5333

ezaquarii:
Fixes #393
Fixes nextcloud/android#5245
Fixes nextcloud/android#5305
Fixes nextcloud/android#5310
Fixes nextcloud/android#5331
Fixes nextcloud/android#5333
Fixes nextcloud/android#5390

### DIFF
--- a/src/main/java/com/nextcloud/common/NextcloudClient.kt
+++ b/src/main/java/com/nextcloud/common/NextcloudClient.kt
@@ -2,6 +2,7 @@
  *
  *   @author Tobias Kaminsky
  *   Copyright (C) 2019 Tobias Kaminsky
+ *   Copyright (C) 2020 Chris Narkiewicz <hello@ezaquarii.com>
  *   Copyright (C) 2019 Nextcloud GmbH
  *   
  *   Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -38,10 +39,7 @@ import com.owncloud.android.lib.common.network.RedirectionPath
 import com.owncloud.android.lib.common.operations.RemoteOperation
 import com.owncloud.android.lib.common.operations.RemoteOperationResult
 import com.owncloud.android.lib.common.utils.Log_OC
-import okhttp3.CookieJar
-import okhttp3.OkHttpClient
-import okhttp3.Request
-import okhttp3.Response
+import okhttp3.*
 import org.apache.commons.httpclient.HttpStatus
 import java.io.IOException
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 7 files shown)


src/main/java/com/nextcloud/common/NextcloudClient.kt
```diff
@@ -50,40 +47,47 @@ import javax.net.ssl.SSLSession
...
+        } catch (ex: Exception) {
+            RemoteOperationResult(ex)
+        }
     }
     
     fun execute(method: OkHttpMethodBase): Int {
         return method.execute(this)
     }
 
+    @Throws(IOException::class)
```
ezaquarii: @tobiasKaminsky This will proably break clients, but since `IOException` is checked, it must be handled.
There is no API to return valid `Response` in case of error, as `OkHttp3` use exceptions to signal abnormal condition. We should follow.

src/main/java/com/nextcloud/common/NextcloudClient.kt
```diff
@@ -50,40 +47,47 @@ import javax.net.ssl.SSLSession
 import javax.net.ssl.TrustManager
 
 
-class NextcloudClient(var baseUri: Uri, val context: Context) {
+class NextcloudClient(var baseUri: Uri, val context: Context, client: OkHttpClient) {
     lateinit var credentials: String
```
ezaquarii: Misused `lateinit`. This should be initialized here. I'll open another ticket for this, tho.

ezaquarii:
@tobiasKaminsky Could you pealse check the method annotated with @Throws? I won't have bandwidth to do it now.

src/main/java/com/nextcloud/common/NextcloudClient.kt
tobiasKaminsky: Hm. I would go instead with try/catch so that we can return our own error code.
After all, this is in our library and also should only be called within the library.

src/main/java/com/nextcloud/common/NextcloudClient.kt
tobiasKaminsky: This is fixed via #396

src/main/java/com/nextcloud/common/NextcloudClient.kt
ezaquarii: I wrapped the exception in a union-like object.

src/main/java/com/nextcloud/common/NextcloudClient.kt
```diff
@@ -50,42 +47,58 @@ import javax.net.ssl.SSLSession
...
     lateinit var userId: String
     lateinit var request: Request
     var followRedirects = true;
-    val client: OkHttpClient
+    val client = client
+
+    data class ResponseOrError private constructor(val result: Response?, val error: Exception?) {
```
tobiasKaminsky: Nice approach :+1: 
To not have too many things in one class, can we extract this into its own class?

ezaquarii:
Go ahead - I don't mind. It would be speedier if you can do it yourself in another commit in this branch as I can't work on it today due to work related schedule 

BTW, while being here, could you please mark internal interfaces with a keyword internal? This would make them unavailable outside library code (whole library, not a package; internal works a bit differently from Java's package private).

The subtle difference is:
Package private - inside the package only.
Internal - access in whole compilation unit, which in this case is nextcloud-android-library project, but not in nextcloud-android app.

By marking things internal, we can safely refactor inside the library boundaries, without risk of breaking library users that might be using our private-but-still-bublic APIs.

Cheers,
Chris

src/main/java/com/nextcloud/common/NextcloudClient.kt
ezaquarii: extracted

ezaquarii:
@tobiasKaminsky I made the `okhttp3` `execute()` request internal. Please confirm if this is what we can do, ie. we don't plan to use it outside the library.

tobiasKaminsky:
> @tobiasKaminsky I made the `okhttp3` `execute()` request internal. Please confirm if this is what we can do, ie. we don't plan to use it outside the library.

Should be fine, I currently cannot imagine why we need this outside our lib, but we can still change it later, if required.

PR #396, opened 2020-02-04
remove lateinit

### LINKED ISSUE #395: NextcloudClient misuses `lateinit` to allow uninitialized variables
All variables marked as `lateinit` should be initialized.
Preferably `lateinit` should be removed.

tobiasKaminsky:
Fix #395 

Based on #394, so let us first merge #394, then I'll rebase this to master, for proper check, etc.
(I just did it this way to have it clean and review can start right away)

### DIFF
--- a/src/androidTest/java/com/owncloud/android/AbstractIT.java
+++ b/src/androidTest/java/com/owncloud/android/AbstractIT.java
@@ -97,9 +97,9 @@ public abstract class AbstractIT {
 
         OwnCloudClientManagerFactory.setUserAgent("Mozilla/5.0 (Android) Nextcloud-android/1.0.0");
 
-        nextcloudClient = new NextcloudClient(url, context);
-        nextcloudClient.credentials = Credentials.basic(loginName, password);
-        nextcloudClient.userId = loginName; // for test same as userId
+        String userId = loginName; // for test same as userId
+        String credentials = Credentials.basic(loginName, password);
+        nextcloudClient = new NextcloudClient(url, userId, credentials, context);
 
         testConnection();
     }
--- a/src/main/java/com/nextcloud/common/NextcloudClient.kt
+++ b/src/main/java/com/nextcloud/common/NextcloudClient.kt
@@ -39,7 +39,9 @@ import com.owncloud.android.lib.common.network.RedirectionPath
 import com.owncloud.android.lib.common.operations.RemoteOperation
 import com.owncloud.android.lib.common.operations.RemoteOperationResult
 import com.owncloud.android.lib.common.utils.Log_OC
-import okhttp3.*
+import okhttp3.CookieJar
+import okhttp3.OkHttpClient
+import okhttp3.Request
 import org.apache.commons.httpclient.HttpStatus
 import java.io.IOException
 import java.util.concurrent.TimeUnit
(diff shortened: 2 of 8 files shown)


src/androidTest/java/com/owncloud/android/AbstractIT.java
```diff
@@ -97,9 +97,9 @@ public static void beforeAll() throws InterruptedException,
 
         OwnCloudClientManagerFactory.setUserAgent("Mozilla/5.0 (Android) Nextcloud-android/1.0.0");
 
-        nextcloudClient = new NextcloudClient(url, context);
-        nextcloudClient.credentials = Credentials.basic(loginName, password);
-        nextcloudClient.userId = loginName; // for test same as userId
+        String userId = loginName; // for test same as userId
+        String credentials = Credentials.basic(loginName, password);
+        nextcloudClient = new NextcloudClient(url, userId, credentials, context);
```
AndyScherzinger: couldn't you just use `loginName`, creating a second variable with the same value and then using both here seems a bit confusing.

src/main/java/com/owncloud/android/lib/common/OwnCloudClientManager.java
```diff
@@ -193,28 +195,36 @@ public NextcloudClient getNextcloudClientFor(OwnCloudAccount account, Context co
...
+
             // TODO v2
             //client.getParams().setCookiePolicy(CookiePolicy.BROWSER_COMPATIBILITY);
             // enable cookie tracking
 
             //AccountUtils.restoreCookies(accountName, client, context);
 
             account.loadCredentials(context);
-            client.credentials = account.getCredentials().toOkHttpCredentials();
+            val credentials = account.getCredentials().toOkHttpCredentials();
```
AndyScherzinger: I'm fine adding `val` via lombok in java code but we should then decide to do so. What do you think @tobiasKaminsky @ezaquari? Just asking since this adds a kotlin language feature to java code. Not sure if we want to mix language concepts. I am not having a strong opinion on this. We should then just document such a rule.

tobiasKaminsky: No, this was just a mistake, as I thought I am writing Kotlin code.
I will change this.

src/androidTest/java/com/owncloud/android/AbstractIT.java
tobiasKaminsky: True, but as this is for test only, I wanted to have it clear that userId and login credentials are normally not the same.

ezaquarii:
Java equivalent of `val` would be `final String` in this case or `final var` (Java 10+).

@AndyScherzinger How can you use `val` in Java using Lombok? I think this feature is not part of Java. Why they even do this kind of stuff?

AndyScherzinger:
@ezaquarii see https://projectlombok.org/features/val - and I don't know why they introduced this feature

tobiasKaminsky:
You can discuss this, if you want ;-)
But I already removed this lombok val and we really should not mix this up…

AndyScherzinger:
It seems to me we are all on the same page, so all good 👍

src/main/java/com/nextcloud/common/NextcloudClient.kt
```diff
@@ -50,48 +50,54 @@ import javax.net.ssl.SSLSession
...
 
-class NextcloudClient(var baseUri: Uri, val context: Context) {
-    lateinit var credentials: String
-    lateinit var userId: String
-    lateinit var request: Request
-    var followRedirects = true;
-    val client: OkHttpClient
-    
+class NextcloudClient(var baseUri: Uri,
+                      var userId: String,
```
ezaquarii: Please note that by default, Kotlin members are `public`, contrary to Java, where default scope is package private.

We shall close the access as much as possible. If we need to keep them accessible insidde tests, `internal` scope can be used to limit access to library only.

tobiasKaminsky:
Test was successful, but codecov timed out, thus merging.

PR #409, opened 2020-03-04
Handle exceptions on new network

tobiasKaminsky:
Another try to fix lots of Exceptions during network call.
Problem is that okhttp throws exceptions when timeout, unknown socket, etc.
We did not catch this so far.


tobiasKaminsky:
Let us try this for 3.11 RC1

PR #452, opened 2020-06-09
Harden CapabilityBooleanType: can never be null

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/5045

### DIFF
--- a/src/main/java/com/owncloud/android/lib/resources/status/CapabilityBooleanType.java
+++ b/src/main/java/com/owncloud/android/lib/resources/status/CapabilityBooleanType.java
@@ -24,27 +24,28 @@
  */
 package com.owncloud.android.lib.resources.status;
 
+import androidx.annotation.NonNull;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 
 /**
  * Enum for Boolean Type in OCCapability parameters, with values:
  * -1 - Unknown
- *  0 - False
- *  1 - True
+ * 0 - False
+ * 1 - True
  */
 @AllArgsConstructor
 public enum CapabilityBooleanType {
-    UNKNOWN (-1),
-    FALSE (0),
-    TRUE (1);
+    UNKNOWN(-1),
+    FALSE(0),
+    TRUE(1);
 
-    @Getter private int value;
+    @Getter
+    private int value;
 
-    public static CapabilityBooleanType fromValue(int value)
-    {
-        switch (value)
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/lib/resources/status/CapabilityBooleanType.java
```diff
@@ -52,7 +54,7 @@ public static CapabilityBooleanType fromValue(int value)
             case 1:
                 return TRUE;
         }
-        return null;
+        return UNKNOWN;
```
AndyScherzinger, line 57: wouldn't this be a standard case for switch/case/default?

tobiasKaminsky: switch refuses to have a default on int. 
maybe I could use <1 and >1, but this is then even stranger…🙈

AndyScherzinger: fine for me 👍 Let's keep it this way

src/main/java/com/owncloud/android/lib/resources/status/CapabilityBooleanType.java
```diff
@@ -24,54 +24,55 @@
...
+        switch (value) {
             case -1:
                 return UNKNOWN;
             case 0:
                 return FALSE;
             case 1:
                 return TRUE;
         }
-        return null;
+        return UNKNOWN;
```
AndyScherzinger, line 56: for the key you could then just go for `UNKNOWN.getValue()` etc. instead of duplicating the int-values here :)

src/main/java/com/owncloud/android/lib/resources/status/CapabilityBooleanType.java
```diff
@@ -24,54 +24,55 @@
...
+        if (boolValue) {
             return TRUE;
         } else {
             return FALSE;
         }
     }
 
-    public boolean isUnknown(){
+    public boolean isUnknown() {
         return getValue() == -1;
```
AndyScherzinger, line 68: same here `UNKNOWN.getValue()` or maybe `UNKNOWN.value`, same for the other checks

AndyScherzinger:
@tobiasKaminsky been nitpicking a bit, feel free to skip my latest comments or apply them :)

PR #483, opened 2020-08-17
Let exceptions in DownloadFileRemoteOperation bubble up

### LINKED ISSUE #482: NPE: RemoteOperationResult - StatusLine.getStatusCode()
When making heavy use of Nextcloud's DocumentsProvider via Seedvault, I sometimes get crashes like these:

This is with Nextcloud `3.12.1` which seems to be using 4b90880c134a7d88929242b2a270dd90c8384d36 of this library.

```ruby
    java.lang.NullPointerException: Attempt to invoke virtual method 'int org.apache.commons.httpclient.StatusLine.getStatusCode()' on a null object reference
        at org.apache.commons.httpclient.HttpMethodBase.getStatusCode(HttpMethodBase.java:570)
        at com.owncloud.android.lib.common.operations.RemoteOperationResult.<init>(RemoteOperationResult.java:328)
        at com.owncloud.android.lib.resources.files.DownloadFileRemoteOperation.run(DownloadFileRemoteOperation.java:85)
        at com.owncloud.android.lib.common.operations.RemoteOperation.execute(RemoteOperation.java:181)
        at com.owncloud.android.operations.DownloadFileOperation.run(DownloadFileOperation.java:160)
        at com.owncloud.android.lib.common.operations.RemoteOperation.execute(RemoteOperation.java:181)
        at com.owncloud.android.files.services.FileDownloader.downloadFile(FileDownloader.java:471)
        at com.owncloud.android.files.services.FileDownloader.access$500(FileDownloader.java:81)
        at com.owncloud.android.files.services.FileDownloader$ServiceHandler.handleMessage(FileDownloader.java:422)
```

It seems like `DownloadFileRemoteOperation#run()` is calling `DownloadFileRemoteOperation#downloadFile()` which swallows all exceptions (not sure if logging works, I didn't see any messages). Then `run()` is working on the potentially bogus result and in this case crashes. I've often seen a status of `-1` returned in the log  without crashes, but that probably also means that there was a problem with the HTTP request.

This crash is especially bad, because the DocumentsProvider consumer is hanging waiting for a response from Nextcloud which never comes.

So if `DownloadFileRemoteOperation#downloadFile()` is only ever called by `run()` which catches all exceptions already, then I suggest to remove the catching from `downloadFile()`, so that they appear in `run()` and this doesn't try to proceed with an undefined result, but errors out properly.

grote:
so they can get handled by the operation's run() method.

Fixes #482 (which has more information)

### DIFF
--- a/src/main/java/com/owncloud/android/lib/resources/files/DownloadFileRemoteOperation.java
+++ b/src/main/java/com/owncloud/android/lib/resources/files/DownloadFileRemoteOperation.java
@@ -100,8 +100,8 @@ public class DownloadFileRemoteOperation extends RemoteOperation {
     }
 
 
-    private int downloadFile(OwnCloudClient client, File targetFile) throws IOException {
-        int status = -1;
+    private int downloadFile(OwnCloudClient client, File targetFile) throws IOException, OperationCancelledException {
+        int status;
         boolean savedFile = false;
         getMethod = new GetMethod(client.getWebdavUri() + WebdavUtils.encodePath(remotePath));
         Iterator<OnDatatransferProgressListener> it;
@@ -175,8 +175,6 @@ public class DownloadFileRemoteOperation extends RemoteOperation {
                 client.exhaustResponse(getMethod.getResponseBodyAsStream());
             }
 
-        } catch (Exception e) {
-          Log_OC.e(TAG, e.getMessage());  
         } finally {
             if (fos != null) fos.close();
             if (!savedFile && targetFile.exists()) {


grote:
Hey looks like this removed a `Dodgy code Warning`. \o/

src/main/java/com/owncloud/android/lib/resources/files/DownloadFileRemoteOperation.java
```diff
@@ -175,8 +175,6 @@ private int downloadFile(OwnCloudClient client, File targetFile) throws IOExcept
                 client.exhaustResponse(getMethod.getResponseBodyAsStream());
             }
 
-        } catch (Exception e) {
```
tobiasKaminsky, line 178: What happens if another Exception is thrown? (if any)

grote: So according to my usage search this private `downloadFile()` method is only ever called from `run()` in the same class. There, the call is already wrapped in a try/catch that catches all `Exception`s, but there it gets handled, while here it just gets swallowed.

tobiasKaminsky: My concern was that any other exception might lead to a crash, but now I see that this is handled then in run(), as you wrote :+1:

tobiasKaminsky (approved):

PR #494, opened 2020-08-27
Add tests for gallery

tobiasKaminsky:

### DIFF
--- a/src/androidTest/java/com/owncloud/android/lib/resources/files/SearchRemoteOperationTest.java
+++ b/src/androidTest/java/com/owncloud/android/lib/resources/files/SearchRemoteOperationTest.java
@@ -332,12 +332,112 @@ public class SearchRemoteOperationTest extends AbstractIT {
     }
 
     @Test
+    public void testGallerySearchLimit() throws IOException {
+        for (int i = 0; i < 10; i++) {
+            String filePath = createFile("image" + i);
+            String remotePath = "/image" + i + ".jpg";
+            assertTrue(new UploadFileRemoteOperation(filePath, remotePath, "image/jpg", "123")
+                    .execute(client).isSuccess());
+        }
+        String videoPath = createFile("video");
+        assertTrue(new UploadFileRemoteOperation(videoPath, "/video.mp4", "video/mpeg", "123")
+                .execute(client).isSuccess());
+
+        String filePath = createFile("pdf");
+        assertTrue(new UploadFileRemoteOperation(filePath, "/pdf.pdf", "application/pdf", "123")
+                .execute(client).isSuccess());
+
+        // get all
+        SearchRemoteOperation sut = new SearchRemoteOperation("image/%", SearchRemoteOperation.SearchType.GALLERY_SEARCH,
+                false);
+
+        RemoteOperationResult result = sut.execute(client);
+        assertTrue(result.isSuccess());
+        assertEquals(11, result.getData().size());
+
+        // limit to 5
+        sut.setLimit(5);
+
+        result = sut.execute(client);
+        assertTrue(result.isSuccess());
+        assertEquals(5, result.getData().size());
+    }
+
+    @Test
+    public void testGallerySearchTimestamp() throws IOException {
+        for (int i = 0; i < 10; i++) {
+            String filePath = createFile("image" + i);
+            String remotePath = "/image" + i + ".jpg";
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
@nextcloud/server-triage  do you have an idea why this is flaky?

https://github.com/nextcloud/android-library/blob/6e1be8b9f327ddff78a8d26547007881cc800cff/src/androidTest/java/com/owncloud/android/lib/resources/files/SearchRemoteOperationTest.java#L228-L242

- It creates 10 images
- finds 10 (always)
- then limit to 5
- and then the query sometimes succeeds, sometimes not

kesselb:
Is there something in nextcloud.log?

tobiasKaminsky:
> Is there something in nextcloud.log?

Nothing. It just seems to work, but does not give the correct amount of items back…

tobiasKaminsky:
@nextcloud/server-triage any idea?

skjnldsv:
I have honestly no idea what to look at :see_no_evil:

tobiasKaminsky:
Here is the test which fails: https://github.com/nextcloud/android-library/pull/494#issuecomment-684832816

So test is very very flaky…

skjnldsv:
Maybe it's asynchronous and the 10 images creation is not done ? :shrug:

tobiasKaminsky:
I fear this is something else…
This is how it behaves:
-    It creates 10 images
-    finds 10 (always)
-    then limit to 5
-    and then the query sometimes succeeds, sometimes not

So only the limited search is wrong…
I ran it multiple times locally against docker ci server:
1) 5 found
2) 1 found
3) 3 found

tobiasKaminsky:
@rullzer @MorrisJobke @icewind1991 any ideas on this? 
I have no idea what is going on, and I do not want to merge this, if it even does not work on a vanilla test server…

MorrisJobke:
After some debugging:

the search in the files code will return any files in the users directory. In my case this included a file inside the trash bin:

Bildschirmfoto 2021-01-20 um 17 03 49


This is then filtered out in the `FilesSearchBackend` of the dav app:

https://github.com/nextcloud/server/blob/cb057829f72c70e819f456edfadbb29d72dba832/apps/dav/lib/Files/FileSearchBackend.php#L161-L173

We then found, that if the limitToHome is set in the searchQuery it will properly do the search in the files code already and does not return other stuff at that level:

https://github.com/nextcloud/server/blob/8b64e92b9262d2a2eec6345685ce421050f95c66/lib/private/Files/Cache/Cache.php#L859-L861

So we added following entry to the where statement:

```xml
                <d:eq>
                    <d:prop>
                        <oc:owner-id xmlns:oc="http://owncloud.org/ns" />
                    </d:prop>
                    <d:literal>test</d:literal>
                </d:eq>
```

so that the limitToHome is properly set in the searchQuery that is created from the DAV search request.

But shouldn't it be like that in the first place? cc @icewind1991 @rullzer

MorrisJobke:
And why isn't the scope not used properly in this case?

```xml
<?xml version="1.0" encoding="UTF-8"?>
<d:searchrequest xmlns:d="DAV:" xmlns:oc="http://nextcloud.com/ns">
    <d:basicsearch>
        <d:select>
            <d:prop>
                <d:getcontenttype />
                <d:resourcetype />
                <d:getcontentlength />
                <d:getlastmodified />
            </d:prop>
        </d:select>
        <d:from>
            <d:scope>
                <d:href>/files/test</d:href>
                <d:depth>infinity</d:depth>
            </d:scope>
        </d:from>
        <d:where>
            <d:and>
                <d:eq>
                    <d:prop>
                        <oc:owner-id xmlns:oc="http://owncloud.org/ns" />
                    </d:prop>
                    <d:literal>test</d:literal>
                </d:eq>
                <d:lt>
                    <d:prop>
                        <d:getlastmodified />
                    </d:prop>
                    <d:literal>1611068822</d:literal>
                </d:lt>
                <d:or>
                    <d:like>
                        <d:prop>
                            <d:getcontenttype />
                        </d:prop>
                        <d:literal>image/%</d:literal>
                    </d:like>
                    <d:like>
                        <d:prop>
                            <d:getcontenttype />
                        </d:prop>
                        <d:literal>video/%</d:literal>
                    </d:like>
                </d:or>
            </d:and>
        </d:where>
        <d:orderby />
        <d:limit>
            <d:nresults>5</d:nresults>
        </d:limit>
    </d:basicsearch>
</d:searchrequest>
````

PR #498, opened 2020-09-02
Make it possible to leave upload retries enabled

grote:
The retries were disabled in 0bfc3857072 and only for upload operations. It is not clear anymore why that happens and if that is still useful. This PR at least gives us the possibility to step by step experiment with retries enabled. The first use of that will be made by the DocumentsProvider's uploads.


tobiasKaminsky:
When you add this flag as an option one, we do not have any drawback, as then the current calls stay the same, but DSP has to explicitly call the new constructor.
This would also prevent all the test changes.

I think this is the way to go as we should not force all users of this lib to adjust their calls, even if they want to have the old behaviour remained.

PR #502, opened 2020-09-04
Make it possible to leave upload retries enabled

grote:
The retries were disabled in 0bfc3857072 and only for upload operations. It is not clear anymore why that happens and if that is still useful. This PR at least gives us the possibility to step by step experiment with retries enabled. The first use of that will be made by the DocumentsProvider's uploads.

### DIFF
--- a/src/main/java/com/owncloud/android/lib/resources/files/UploadFileRemoteOperation.java
+++ b/src/main/java/com/owncloud/android/lib/resources/files/UploadFileRemoteOperation.java
@@ -61,6 +61,7 @@ public class UploadFileRemoteOperation extends RemoteOperation {
 	protected String remotePath;
 	protected String mimeType;
 	private String lastModificationTimestamp;
+	protected final boolean disableRetries;
 	PutMethod putMethod = null;
 	private String requiredEtag = null;
     String token = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/lib/resources/files/UploadFileRemoteOperation.java
```diff
@@ -79,10 +80,27 @@ public UploadFileRemoteOperation(String localPath,
         this(localPath, remotePath, mimeType, requiredEtag, lastModificationTimestamp, null);
     }
 
+    public UploadFileRemoteOperation(String localPath,
+                                     String remotePath,
+                                     String mimeType,
+                                     String requiredEtag,
+                                     String lastModificationTimestamp,
+									 boolean disableRetries) {
```
tobiasKaminsky, line 88: Somehow there are too many tabs, or?

tobiasKaminsky (approved):

src/main/java/com/owncloud/android/lib/resources/files/UploadFileRemoteOperation.java
grote: The formatting in this library is a bit of a mess (mix of tabs and spaces) and I didn't want to auto-format everything. The above parameters seem to use spaces instead of tabs. I moved the new one to spaces as well.

tobiasKaminsky (dismissed):

PR #503, opened 2020-09-08
prefer IPV6 addresses if available

200Puls:
I am a bit unsure if the NextcloudClient client is used.
The change that really fixes #3193 is the removal of the code in AdvancedSslSocketFactory.java which used to explicitly prefer IPV4. I'm not sure why this code existed in this form.

### DIFF
--- a/src/main/java/com/nextcloud/common/IPV6PreferringDNS.java
+++ b/src/main/java/com/nextcloud/common/IPV6PreferringDNS.java
@@ -1,0 +1,41 @@
+package com.nextcloud.common;
+
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import okhttp3.Dns;
+
+/**
+ * Implementation of DNS which prefers IPV6 addresses (if available) over IPV4.
+ */
+public class IPV6PreferringDNS implements Dns {
+
+        private Map<String, List<InetAddress>> cache = new HashMap<>();
+
+        @Override
+        public List<InetAddress> lookup(String hostname) throws UnknownHostException {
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
> I am a bit unsure if the NextcloudClient client is used.

This our new client, which will replace OwnCloudClient in future.



>  I'm not sure why this code existed in this form.

Me neither, maybe it was back then to block all ipv6?

src/main/java/com/nextcloud/common/IPV6PreferringDNS.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Implementation of DNS which prefers IPV6 addresses (if available) over IPV4.
+ */
+public class IPV6PreferringDNS implements Dns {
+
+        private Map<String, List<InetAddress>> cache = new HashMap<>();
+
+        @Override
+        public List<InetAddress> lookup(String hostname) throws UnknownHostException {
+            List<InetAddress> addresses = cache.get(hostname.toLowerCase());
```
tobiasKaminsky, line 22:
```suggestion
            List<InetAddress> addresses = cache.get(hostname.toLowerCase(Locale.ROOT));
```

src/main/java/com/owncloud/android/lib/common/network/AdvancedSslSocketFactory.java
```diff
@@ -190,16 +190,6 @@ public Socket createSocket(final String host, final int port,
 
     private InetAddress getInetAddressForHost(String host) throws UnknownHostException {
         InetAddress address = InetAddress.getByName(host);
```
tobiasKaminsky, line 192:
```suggestion
        return InetAddress.getByName(host);
```

tobiasKaminsky (approved):

tobiasKaminsky (approved):

tobiasKaminsky:
Let us closely monitor if users will have problems with it.
It will be in tomorrows dev version

PR #514, opened 2020-10-03
Fix issues with HTTP redirects and file upload

z3ntu:
Firstly the test setup:
* Nextcloud running behind a reverse proxy (e.g. Traefik) configured for HTTP only
* Configure Nextcloud app to connect to that server
* Configure https on reverse proxy, enable http -> https redirects on the reverse proxy
  * Traefik uses the HTTP 308 status code to redirect clients
* Try to upload some pictures via the Auto upload feature

Previously this resulted in the following notification being sent to the user:
Screenshot_20201003-185029_2

With these patches I can successfully upload pictures to my server.

See commit messages for more details.


tobiasKaminsky:
Sorry for coming back this late…
Permanent redirects should already be handled via login, so that the real redirected url is stored.
This way we do not have to do multiple hops per each request.

Have you changed your system after installing the app?

z3ntu:
Yes, as mentioned in the original message, the permanent redirect was added after the initial login leading to the issues I was seeing.

tobiasKaminsky:
Hm. But this then means, that every request is done to times: first against old, get redirect, then to new url.
As the library does not store the base url, this should be handled on app side, to update it, once it sees a permanet redirect.

Do you have an idea how to inform the client?

z3ntu:
Right.. But as non-HTTP 308 redirect statuses already redirect properly I don't think solving this is in the scope of this - and this PR is just adjusting existing redirect code (for the most part).

Also I'm too unfamiliar with the Nextcloud App codebase with regard to updating the base url.

tobiasKaminsky:
I will have a deeper look into this soon.
Currently I am a bit too busy, but I will not forget your PR.

z3ntu:
> but I will not forget your PR.

Safe to say this was forgotten.

I'll leave this open but my setup has changed in the meantime so I'm not directly interested in this anymore.

z3ntu:
@tobiasKaminsky No clue if this is still relevant in 2025, shall I close?

PR #532, opened 2020-11-02
Allow method on plain client

tobiasKaminsky:

### DIFF
--- a/src/androidTest/java/com/owncloud/android/lib/resources/users/StatusIT.kt
+++ b/src/androidTest/java/com/owncloud/android/lib/resources/users/StatusIT.kt
@@ -32,6 +32,11 @@ import org.junit.Before
 import org.junit.Test
 
 class StatusIT : AbstractIT() {
+    companion object {
+        val SECOND_IN_MILLIS = 1000
+        val HOUR_IN_MINUTES = 3600
+    }
+
     @Before
     fun before() {
         val result = GetCapabilitiesRemoteOperation().execute(client)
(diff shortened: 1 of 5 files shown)


src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
```diff
@@ -144,11 +147,31 @@ abstract class OkHttpMethodBase(
...
+        requestHeaders.forEach { (name, value) -> temp.header(name, value) }
+
+        applyType(temp)
+
+        val request = temp.build()
+
+        try {
+            response = client.newCall(request).execute()
+        } catch (ex: IOException) {
+            System.out.println(ex.message)
```
AndyScherzinger, line 169: should be replaced by proper log statement

AndyScherzinger (dismissed):

tobiasKaminsky:
> approved these changes

There is no point in approving/review a "1. developing", as this only gives false positive (e.g. you might not re-check after I switch it to 2. to review and then your valuable feedback is missing).

In this case added sys out log on purpose, to get the real error message:
> java.lang.RuntimeException: javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.
at com.nextcloud.common.OkHttpMethodBase.execute(OkHttpMethodBase.kt:170)

I will fix it tomorrow.

src/androidTest/java/com/owncloud/android/lib/resources/users/StatusIT.kt
```diff
@@ -114,7 +119,7 @@ class StatusIT : AbstractIT() {
         val statusesList: ArrayList<PredefinedStatus> =
             result.singleData as ArrayList<PredefinedStatus>
         val newCustomStatusMessage = statusesList[2]
-        val clearAt = System.currentTimeMillis() / 1000 + 3600 // in one hour
+        val clearAt = System.currentTimeMillis() / SECOND_IN_MILLIS + ONE_HOUR_IN_MINUTES
```
AndyScherzinger, line 122: either drop the ONE_ or also prefix the SECOND_IN_MILLIS :)

AndyScherzinger (approved):
LGTM 👍

PR #534, opened 2020-11-02
Throw RuntimeException when accessing data if operation failed

tobiasKaminsky:

### DIFF
--- a/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
+++ b/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
@@ -482,10 +482,16 @@ public class RemoteOperationResult implements Serializable {
     }
 
     public ArrayList<Object> getData() {
+        if (!mSuccess) {
+            throw new RuntimeException("Accessing result data after operation failed!");
+        }
         return mData;
     }
 
     public Object getSingleData() {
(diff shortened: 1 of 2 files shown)


AndyScherzinger (approved):

AndyScherzinger (approved):

src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
```diff
@@ -482,10 +482,16 @@ public void setSingleData(Object object) {
     }
 
     public ArrayList<Object> getData() {
+        if (!mSuccess) {
+            throw new RuntimeException("Accessing result data after operation failed!");
```
AndyScherzinger, line 486: Not sure if it is worth the effort but throwing a top-level exception is rather uncommon, not sure if there is an existing, better-fitting Runtime child exception or if one should be created (while it should of course be a RuntimeException)

tobiasKaminsky: As this normally would not happen, I'd like to skip the additional effort.

AndyScherzinger: sure thing!

PR #547, opened 2020-11-25
Add content length function

tobiasKaminsky:
Needed for https://github.com/nextcloud/android/pull/7207

### DIFF
--- a/src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
+++ b/src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
@@ -95,6 +95,10 @@ abstract class OkHttpMethodBase(
         return response?.body()?.string() ?: ""
     }
 
+    fun getResponseContentLength(): Long {
+        return response?.body()?.contentLength() ?: -1
+    }
+
     fun releaseConnection() {
         response?.body()?.close()
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
```diff
@@ -115,7 +119,7 @@ abstract class OkHttpMethodBase(
         return response?.header(name)
     }
 
-    fun getRequestHeader(name: String): String? {
```
ezaquarii, line 118: Having a header, having it set to `""` or not at all might make a difference for some clients.
If the client is consuming it in Kotlin, one can use `getRequestHeader("dodgy") :? ""` instead.

That's a big if. Java is probably less fun.

ezaquarii (approved):
One minor thing to double-think before merge, but otherwise looks good to me.
.

src/main/java/com/nextcloud/common/OkHttpMethodBase.kt
tobiasKaminsky: Return of this function was not touched, so it always returns either header or "", therefore String? is unnecessary. 

If we want to make sure that it is a different handling if no header is found, then we need to return either null or something else.
As it is "our" implementation, I am fine with keeping it like that: either return header or empty (aka. not found/set).

PR #549, opened 2020-11-30
Unified search

tobiasKaminsky:


AndyScherzinger (dismissed):
looking good, just some warnigns from the linter/detekt

tobiasKaminsky:
> java.lang.AssertionError: expected:<51> but was:<null>

on Stable 20 we do not yet have the info about fileId.
So I fear that we need to have code for <22 to no use fileId in this case but use remote path (on files PR).

AlvaroBrey (approved):

PR #557, opened 2020-12-18
Use okhttp 4.x branch and dav4jvm-2.x

tobiasKaminsky:
Since we will soon use 5+ only, we can switch to latest branches.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,10 +11,6 @@ updates:
   - 2. to review
   - dependencies
   ignore:
-  - dependency-name: com.gitlab.bitfireAT:dav4jvm
-    versions:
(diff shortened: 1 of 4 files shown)


tobiasKaminsky:
So a quick smoke test with real client was ok.
@AndyScherzinger I would like to merge it as soon as we branch off for 3.15, so that it can be tested in entire dev cycle for 3.16.

AndyScherzinger:
> @AndyScherzinger I would like to merge it as soon as we branch off for 3.15, so that it can be tested in entire dev cycle for 3.16.
@tobiasKaminsky sounds great, yes 👍

AndyScherzinger (dismissed):

AndyScherzinger (approved):

PR #619, opened 2021-05-11
Activity: preview object now has filename

tobiasKaminsky:
made parser more robust

Introduced in https://github.com/nextcloud/activity/commit/38a31e599e0df883c517d76886af38e9babfb140#diff-76b5a357391276b282a516f54f48ef3c207f46d8192dc58c208d5183d38415f8

@nickvergessen do we really want to change api this silently?

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -6,15 +6,6 @@
           <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
         </value>
       </option>
-      <option name="PACKAGES_IMPORT_LAYOUT">
-        <value>
-          <package name="" alias="false" withSubpackages="true" />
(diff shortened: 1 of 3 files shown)


AndyScherzinger (dismissed):

tobiasKaminsky:
> Sorry I was not aware that fullbackwards compatible changes which just add additional fields would break anything.

I guess this heavily depends on the way one writes the parser.
I tried it now to be more graceful.

tobiasKaminsky:
When will the new activity version be out in the wild?

nickvergessen:
https://github.com/nextcloud/activity/pull/578

Seems like it was backported to 21.0.2 which is coming this or next week iirc

szaimen:
Can this be backported to stable-2.6?

PR #622, opened 2021-05-13
Raise CHUNK_SIZEs

szaimen:
fix https://github.com/nextcloud/android/issues/8398
Signed-off-by: szaimen <szaimen@e.mail.de>


tobiasKaminsky:
We will not change the default without any deeper knowledge if this is a good idea.
For other apps that use the library we could introduce a more flexible way, so that the apps can pass their desired chunk size.

PR #635, opened 2021-05-28
Verbose output to debug NPE

tobiasKaminsky:
This is not for merging, but for finding cause of https://github.com/nextcloud/android/issues/8370

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #638, opened 2021-06-04
Fix for failed logins / app crashes

mightymop:
fix https://github.com/nextcloud/android/issues/8302
fix https://github.com/nextcloud/android/issues/8525
fix https://github.com/nextcloud/android/issues/8297
fix https://github.com/nextcloud/android/issues/8370


tobiasKaminsky:
Can you give us some more info why this will fix the problem?
I had created something similar https://github.com/nextcloud/android-library/pull/635 to debug https://github.com/nextcloud/android/issues/8370 …

mightymop:
> Can you give us some more info why this will fix the problem?
> I had created something similar #635 to debug [nextcloud/android#8370](https://github.com/nextcloud/android/issues/8370) …

First i took a look into the corresponding source of the owncloud android client and realized that they are using the blocking method instead of the peek... Then i merged the peace of code into my nextcloud repo and did some tests... and could not reproduce the issue ... furthermore i have added some errorhandling to be sure that the code will not try to auth with a null password which would produce the ldap lock...

mightymop:
Are there any news on the related issue(s)? @tobiasKaminsky You mentioned that you had a similar issue? Could you fix it?

AlvaroBrey (approved):
This is the right approach. As `createNextcloudClient` is not called from the main thread, `blockingGetAuthToken` should be a safe call.

PR #651, opened 2021-06-30
Retrieve localId via webdav

tobiasKaminsky:
Needed for https://github.com/nextcloud/android/pull/8643

### DIFF
--- a/library/src/androidTest/java/com/nextcloud/android/lib/resources/search/UnifiedSearchRemoteOperationIT.kt
+++ b/library/src/androidTest/java/com/nextcloud/android/lib/resources/search/UnifiedSearchRemoteOperationIT.kt
@@ -75,7 +75,7 @@ class UnifiedSearchRemoteOperationIT : AbstractIT() {
 
         assertNotNull(firstResult)
         assertEquals(remotePath, firstResult?.remotePath())
-        assertEquals(fileId, firstResult?.fileId())
+        assertEquals(fileId.toString(), firstResult?.fileId())
     }
 
     @Test
--- a/library/src/androidTest/java/com/owncloud/android/FileIT.java
+++ b/library/src/androidTest/java/com/owncloud/android/FileIT.java
@@ -328,4 +328,21 @@ public class FileIT extends AbstractIT {
             assertTrue(sharees.contains(user));
         }
     }
+
+    @Test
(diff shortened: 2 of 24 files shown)


src/main/java/com/owncloud/android/lib/resources/activities/GetActivitiesRemoteOperation.java
```diff
@@ -102,7 +102,7 @@ public RemoteOperationResult run(NextcloudClient client) {
         String url = client.getBaseUri() + OCS_ROUTE_V12_AND_UP;
 
         // add filter for fileId, if available
-        if (!fileId.isEmpty()) {
+        if (fileId > 0) {
```
AndyScherzinger, line 105: "unitialized" value is actually ```-1``` so is 0 considered "empty" and thus the ```>0``` ?

tobiasKaminsky: True, but fileId can never be 0 on server.
-1 is common for uninitialized.
But you are right that this is a bit strange.

AndyScherzinger (approved):

AndyScherzinger (approved):

AndyScherzinger (approved):

AlvaroBrey (dismissed):

tobiasKaminsky:
@AlvaroBrey 
```
java.lang.AssertionError: expected: java.lang.Long<69> but was: java.lang.String<69>
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.failNotEquals(Assert.java:835)
at org.junit.Assert.assertEquals(Assert.java:120)
at org.junit.Assert.assertEquals(Assert.java:146)
at com.nextcloud.android.lib.resources.search.UnifiedSearchRemoteOperationIT.filesSearch(UnifiedSearchRemoteOperationIT.kt:78)
```
Attributes is Map<String, String>. 
Thus I dare not to touch it.

I now fixed the test, by converting long to String.

AlvaroBrey (approved):

AlvaroBrey (dismissed):

AlvaroBrey (approved):

AlvaroBrey (dismissed):

AlvaroBrey (approved):

PR #663, opened 2021-07-07
delete outdated comments

tobiasKaminsky:
I added 4 comments which simulate failing tests.
Those should then be removed after successful drone.


AndyScherzinger (approved):
LGTM

PR #699, opened 2021-08-24
Explicit specify jdk 11 for jitpack

tobiasKaminsky:


build.gradle
```diff
@@ -197,3 +199,16 @@ detekt {
...
 }
+
+afterEvaluate {
+    publishing {
+        publications {
+            release(MavenPublication) {
+                from components.release
+
+                groupId = 'com.nextcloud.android-library'
+                artifactId = 'master'
```
AndyScherzinger, line 210: Does this lock builds for master or will everything still work for tags and branches?

tobiasKaminsky: This is only the name of the generated aar file, which is then renamed by jitpack.
```
dependencies {
	        implementation 'com.github.nextcloud:android-library:jitpack-SNAPSHOT'
	}
```
Will work, with this branch.

PR #707, opened 2021-09-02
Revert dav4jvm update for now as it causes problems on testing in files app

tobiasKaminsky:


tobiasKaminsky:
This is not a good solution, but for now it at least helps us to work again.
After this merge we can create a new PR with the version bump again to use this for finding/fixing root cause.

tobiasKaminsky:
Closing in favor of excluding "junit" within files app:
```
    implementation ("com.github.nextcloud:android-library:$androidLibraryVersion") {
        exclude group: "junit"
    }
```

Issue #696, opened 2021-08-21
Timeout when assembling large file uploads


minfaer:
As reported to the android-app repository in nextcloud/android#5609, uploading large files fails at 100% by timeout if the server takes too long assembling them. This results in an endless loop of retries and conflicts.

This can be fixed by increasing the read-timeout of the move-method that does the assembly in the chunk-upload. I tested setting the timeout to a flat 15 minutes similarly to what was done to version 1 of the library in #19. However, maybe that timeout should scale with the filesize only for files larger than some limit (like 100mb) and be left alone otherwise?

If desired, I can submit a pr for either solution.

minfaer:
As the question came up in the aforementioned issue: the desktop-client adjusts the time-out depending on filesize in [this function](https://github.com/nextcloud/desktop/blob/fe65715b975b3b3865fbc06151337072aa53467f/src/libsync/propagateupload.cpp#L685)
to 3 minutes per GB

tobiasKaminsky:
Thanks for this hint!

gomme600:
Adjusting the timeout seems like the best idea. This issue has been wreaking havok for quite a while for my videos uploads. I have almost 300gb of duplicate files!

PR #709, opened 2021-09-04
Scale read-timeout on chunk-upload file assembly

minfaer:
This is a fix for nextcloud/android#5609 as proposed in #696. 
It works by setting the read timeout for the move-method that assembles chunk-uploads on the server to 3 minutes per GB between 30 seconds and 30 minutes, consistent with Nextcloud-dektop.

### DIFF
--- a/src/main/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperation.java
+++ b/src/main/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperation.java
@@ -50,6 +50,7 @@ import java.nio.channels.FileChannel;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.lang.Math;
 
 
 public class ChunkedFileUploadRemoteOperation extends UploadFileRemoteOperation {
@@ -181,7 +182,7 @@ public class ChunkedFileUploadRemoteOperation extends UploadFileRemoteOperation 
             if (token != null) {
                 moveMethod.addRequestHeader(E2E_TOKEN, token);
             }
-            int moveResult = client.executeMethod(moveMethod);
+            int moveResult = client.executeMethod(moveMethod, CalcAssembleTimeout(file), -1);
 
             result = new RemoteOperationResult(isSuccess(moveResult), moveMethod);
         } catch (Exception e) {
@@ -314,5 +315,13 @@ public class ChunkedFileUploadRemoteOperation extends UploadFileRemoteOperation 
         }
 
         return putMethod;
+    }
+
+    private int CalcAssembleTimeout(File file) {
+        final double threeMinutes = 3.0 * 60 * 1000;
+        final int thirtySeconds = 30 * 1000;
+        final int thirtyMinutes = 30 * 60 * 1000;
+        int AssembleReadTimeout = Math.max(thirtySeconds , Math.min((int)(threeMinutes * file.length() / 1e9) , thirtyMinutes) );
+        return AssembleReadTimeout;
     }
 }


src/main/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperation.java
```diff
@@ -315,4 +316,12 @@ private PutMethod createPutMethod(String uriPrefix) {
 
         return putMethod;
     }
+
+    private int CalcAssembleTimeout(File file) {
```
tobiasKaminsky, line 320:
```suggestion
    private int calcAssembleTimeout(File file) {
```

src/main/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperation.java
```diff
@@ -181,7 +182,7 @@ protected RemoteOperationResult run(OwnCloudClient client) {
             if (token != null) {
                 moveMethod.addRequestHeader(E2E_TOKEN, token);
             }
-            int moveResult = client.executeMethod(moveMethod);
+            int moveResult = client.executeMethod(moveMethod, CalcAssembleTimeout(file), -1);
```
tobiasKaminsky, line 185: For -1 can you introduce a const, e.g. DO_NOT_CHANGE_DEFAULT, so that when reading the code one directly knows that this value does not change connection timeout.

src/main/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperation.java
```diff
@@ -315,4 +316,12 @@ private PutMethod createPutMethod(String uriPrefix) {
 
         return putMethod;
     }
+
+    private int CalcAssembleTimeout(File file) {
+        final double threeMinutes = 3.0 * 60 * 1000;
+        final int thirtySeconds = 30 * 1000;
+        final int thirtyMinutes = 30 * 60 * 1000;
+        int AssembleReadTimeout = Math.max(thirtySeconds , Math.min((int)(threeMinutes * file.length() / 1e9) , thirtyMinutes) );
```
tobiasKaminsky, line 324:
```suggestion
return Math.max(thirtySeconds , Math.min((int)(threeMinutes * file.length() / 1e9) , thirtyMinutes) );
```

tobiasKaminsky:
@minfaer thanks for your PR :tada: 
If you have time it would be really great if you can write some unit tests in https://github.com/nextcloud/android-library/blob/fba3a49e911c369771bfdf0013b97e10e6979157/src/test/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperationTest.java#L73

Different file sizes:
- 0b
- 100b
- 100 Mb
- 1 Gb
- 2 Gb
- 50 Gb
- 500 Gb

If you do not have time, then we will take care of it :)

src/main/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperation.java
```diff
@@ -315,4 +316,12 @@ private PutMethod createPutMethod(String uriPrefix) {
 
         return putMethod;
     }
+
+    private int CalcAssembleTimeout(File file) {
+        final double threeMinutes = 3.0 * 60 * 1000;
```
AlvaroBrey, line 321: I don't think this needs to be a `double`, an `int` should work fine.

src/main/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperation.java
AlvaroBrey: I'd also suggest extracting `file.length() / 1e9` into a new variable named `fileSizeInGb` to make it easier to read and make the algorithm clearer.

AlvaroBrey:
Sorry for the close and reopen, my fingers slipped :facepalm:   I've added a couple of minor comments. Thanks for your contribution!

minfaer:
Thank You @tobiasKaminsky and @AlvaroBrey for Your feedback! I implemented your suggestionsand am looking into writing the unit-tests.

src/main/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperation.java
minfaer: In the original version, this was a double to prevent `threeMinutes * file.length() / 1e9` from truncating to full Gb, but by declaring `fileSizeInGb` as double, this has indeed become unnecessary.

m-terlinde:
Thanks @minfaer for this suggested patch!
Since this issue is bugging me for month, I'd love to see the patch in production.

tobiasKaminsky:
I added tests and once CI is passed, we can merge it.
Thanks @minfaer

PR #727, opened 2021-10-13
scripts: lint-up.rb: Have Drone signoff the commits so it doesn't break DCO

AlvaroBrey:


tobiasKaminsky:
> We could also just set it ourselves for the command with the [`GIT_COMMITTER_EMAIL`](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_committing) variable, if there is no better way of doing it.

Can you try this as a short term solution, but I would also like to have a general solution :)

tobiasKaminsky:
Cool, can you rebase and remove all the testing commits, so that we only have the one change :)

PR #744, opened 2021-11-03
Do not check dependencies as we have DependAbot for this

tobiasKaminsky:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -127,6 +127,7 @@ android {
         abortOnError false
         htmlReport true
         htmlOutput file("$project.buildDir/reports/lint/lint.html")
+        disable 'GradleDependency'
     }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -124,9 +124,11 @@ android {
     }
 
     lintOptions {
+        checkGeneratedSources = true
```
AlvaroBrey, line 127: Do we want `checkGeneratedSources`? What is it used for?

tobiasKaminsky: c&p from files: https://github.com/nextcloud/android/blob/master/build.gradle#L106

This is for e.g. dagger, lombok, etc.
Indeed I am unsure if this is good at all, I will remove it here.

PR #749, opened 2021-11-05
As lint 0 🎉 we do not need lint-up anymore, but can rely on plain lint GH action

tobiasKaminsky:
- [x] ./gradlew lint does not fail if there is a warning


AlvaroBrey:
Spotbugs file is missing for some reason

AlvaroBrey (approved):

PR #756, opened 2021-11-15
encode user id in `OwnCloudClient.getFilesDavUri(String)`

mutec:
usernames containing spaces cause serious issues up to denying logins and syncs - making the whole app unusable while the same id works using the web gui.

Fixes #755

PR #761, opened 2021-11-23
Fix username encoding in file paths

AlvaroBrey:
https://github.com/nextcloud/android-library/pull/756 introduced encoding for username as an attempt to fix some issues for usernames with spaces.
However that backfired spectacularly, causing even more issues because encoding was now inconsistent throughout the code.
This fixes that.

Fixes https://github.com/nextcloud/android/issues/9331

### DIFF
--- a/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
+++ b/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
@@ -301,7 +301,7 @@ public class OwnCloudClient extends HttpClient {
     }
 
     public Uri getFilesDavUri() {
-        return Uri.parse(getDavUri() + "/files/" + userId);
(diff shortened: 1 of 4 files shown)


tobiasKaminsky (approved):

tobiasKaminsky:
@AlvaroBrey what about adding an user with whitespaces to drone and use this one then as default for testing?
Shall I quickly do this?

AlvaroBrey:
> @AlvaroBrey what about adding an user with whitespaces to drone and use this one then as default for testing? Shall I quickly do this?

I don't see why not, if an user with spaces works, one without spaces should work too

AlvaroBrey:
Merging this as we need to trigger a build on Jitpack to see if the dependency errors persist, as it's messing up all our CI. We can implement the drone changes later.

PR #762, opened 2021-11-24
SearchResultEntry: detect file results not from Files provider

AlvaroBrey:
Some providers like FullTextSearch give file results without the `fileId` attribute.
This refactor allows for that, otherwise they would crash the UnifiedSearch UI.

Fixes https://github.com/nextcloud/android/issues/9345

### DIFF
--- a/src/main/java/com/owncloud/android/lib/common/SearchResultEntry.kt
+++ b/src/main/java/com/owncloud/android/lib/common/SearchResultEntry.kt
@@ -27,7 +27,6 @@
 package com.owncloud.android.lib.common
 
 import android.net.UrlQuerySanitizer
-import java.net.URL
 
 /**
  * One search result entry of an unified search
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/lib/common/SearchResultEntry.kt
```diff
@@ -54,17 +59,20 @@ data class SearchResultEntry(
...
-        val sanitizer = UrlQuerySanitizer()
-        sanitizer.allowUnregisteredParamaters = true
-        sanitizer.unregisteredParameterValueSanitizer = UrlQuerySanitizer.getAllButNulLegal()
+        val sanitizer = UrlQuerySanitizer().apply {
+            allowUnregisteredParamaters = true
+            unregisteredParameterValueSanitizer = UrlQuerySanitizer.getAllButNulLegal()
+        }
+
         sanitizer.parseUrl(resourceUrl)
-        URL(resourceUrl).query?.let { sanitizer.parseQuery(it) }
```
AlvaroBrey, line 61: We don't need to do this. `parseUrl` already calls `parseQuery` if present. Additionally this crashes when trying to parse paths that are not valid URLs (like files paths without the server).

tobiasKaminsky (approved):

tobiasKaminsky:
> Some providers like FullTextSearch give file results without the `fileId` attribute.

If we need this, please create an issue in these repositories, so this will then work automatically at some point.
👏 for having this as a fallback 👏

AlvaroBrey:
> If we need this, please create an issue in these repositories, so this will then work automatically at some point. clap for having this as a fallback clap

Done: https://github.com/nextcloud/files_fulltextsearch/issues/143

PR #782, opened 2021-12-14
allow to query photos and videos also from external storages in gallery

szaimen:
Attempt to fix https://github.com/nextcloud/android/issues/8618

Inspired by https://github.com/nextcloud/photos/pull/804

Best reviewed like this: https://github.com/nextcloud/android-library/pull/782/files?diff=unified&w=1

Signed-off-by: szaimen <szaimen@e.mail.de>


szaimen:
@tobiasKaminsky @AlvaroBrey I have no idea what I am doing so please review carefully :)

src/main/java/com/owncloud/android/lib/resources/files/NcSearchMethod.java
```diff
@@ -287,17 +287,7 @@ private Document createQuery(String searchQuery) {
...
-                Element lastModified = query.createElementNS(NAMESPACE_OC, "oc:owner-id");
-                Element literal = query.createElementNS(DAV_NAMESPACE, "d:literal");
-                Element prop = query.createElementNS(DAV_NAMESPACE, "d:prop");
-                prop.appendChild(lastModified);
-                literal.setTextContent(String.valueOf(userId));
 
-                lessThan.appendChild(prop);
-                lessThan.appendChild(literal);
-
-                and.appendChild(lessThan);
```
szaimen, line 300: We should only delete those lines starting from NC22 and above. Is there a way to do this? (E.g. get the server version at this place and add a condition...

AlvaroBrey: Not directly inside `NcSearchMethod` as far as I know. However `NcSearchMethod` is called only in [`SearchOperation`](http://github.com/nextcloud/android-library/blob/ff59951f63124d2ed7777d312f9b04558f1eb991/src/main/java/com/owncloud/android/lib/resources/files/SearchRemoteOperation.java#L98-L98), so theoretically you could call `GetCapabilitiesRemoteOperation`inside that, and pass the resulting version to `NcSearchMethod`.

Example of calling `GetCapabilitiesRemoteOperation` inside another operation: http://github.com/nextcloud/android-library/blob/ff59951f63124d2ed7777d312f9b04558f1eb991/src/main/java/com/owncloud/android/lib/common/accounts/ExternalLinksOperation.java#L79-L79. You should then be able to use `capability.getVersion().isNewerOrEqual(NextcloudVersion.nextcloud_22)` to check the server version.

AlvaroBrey: By the way, I think you should also probably skip the creation of `and`, and just append `equalsElement` to `whereElement`, like in the `else` block.

szaimen: Thanks for the review!
I hope I fixed everything in https://github.com/nextcloud/android-library/pull/782/commits/c0d96261d18b23e78ec120729b18f2e365f84274
But I must admit that I still understand the syntax only partially.

szaimen: seems like the compiling has failed. What did I do wrong?

tobiasKaminsky: Calling get remote capability all the time is a bit overhead, I think.
Just add the capability to SearchRemoteOperation, so that it is required.

AlvaroBrey (changes requested):
Getting closer now :)

To fix compilation issues:
- Import `GetCapabilitiesRemoteOperation` and `OCCapability` in `SearchRemoteOperation`
- Add the `@JvmField` annotation to `NextcloudVersion.nexctloud_22`, as it is for `nextcloud_22`, otherwise it's private.

Additionally, I think it would be better to pass either the entire `OCCapability` object, or the version, to `NcSearchMethod`, and then check the version inside that (rather than checking the version in `SearchOperation`). This way is more flexible and `SearchOperation` is kept a bit more decoupled from the inner workings of `NcSearchMethod`.

tobiasKaminsky:
> Calling get remote capability all the time is a bit overhead, I think.
Just add the capability to SearchRemoteOperation, so that it is required.

I think I added this to an already resolved communication.

szaimen:
I hope I addressed all points. Though the compilation still fails :/
Please have another look! :)

AlvaroBrey (approved):
Looks good, however this now needs changes in the app, in order to pass the new `OCCapability` param to `SearchRemoteOperation` wherever needed.

tobiasKaminsky:
Client PR in https://github.com/nextcloud/android/pull/9651

tobiasKaminsky (approved):

PR #809, opened 2022-02-01
RemoteOperationResult: don't ignore authentication headers after the first one

AlvaroBrey:
This was causing auth method detection issues in the Files app after the order of headers changed, possibly due to library updates, in 3.18+.

Note: the relevant change is [removing this `break` in the constructor](https://github.com/nextcloud/android-library/pull/809/files#diff-47129d2fe9c1b2a8f38462f0c3cf42d9a11a39807dfbef165ffbfd0fe9ba6b14L386), the rest is just cleaning up a bit.

Fixes https://github.com/nextcloud/android/issues/9486

### DIFF
--- a/src/test/java/com/owncloud/android/lib/common/operations/RemoteOperationResultTest.kt
+++ b/src/test/java/com/owncloud/android/lib/common/operations/RemoteOperationResultTest.kt
@@ -1,0 +1,123 @@
+/*
+ * Nextcloud Android Library is available under MIT license
+ *
+ * @author Álvaro Brey Vilas
+ * Copyright (C) 2022 Álvaro Brey Vilas
+ * Copyright (C) 2022 Nextcloud GmbH
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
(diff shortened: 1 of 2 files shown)


AlvaroBrey:
Note: the relevant change is [removing this `break` in the constructor](https://github.com/nextcloud/android-library/pull/809/files#diff-47129d2fe9c1b2a8f38462f0c3cf42d9a11a39807dfbef165ffbfd0fe9ba6b14L386), the rest is just cleaning up a bit.

tobiasKaminsky:
Nice finding!
As this is our library and we should prevent such a problem, can you add a test please?

AlvaroBrey:
> Nice finding! 

Big credits to @quenenni who provided a test account. No way I could have debugged this otherwise.

>As this is our library and we should prevent such a problem, can you add a test please?
Will do!

src/test/java/com/owncloud/android/lib/common/operations/RemoteOperationResultTest.kt
```diff
@@ -0,0 +1,96 @@
+package com.owncloud.android.lib.common.operations
```
tobiasKaminsky, line 1: Licence please :)
There is even an automated way with templates within AS, if you like such things :)

tobiasKaminsky (approved):

src/test/java/com/owncloud/android/lib/common/operations/RemoteOperationResultTest.kt
AlvaroBrey: Fixed, I have the template setup for Files but forgot for the library :(

tobiasKaminsky (approved):

PR #820, opened 2022-02-17
RemoteOperationResult: ignore location header if there are authentication headers present

AlvaroBrey:
In https://github.com/nextcloud/android-library/commit/d1ace843b510a0788794cd40ce75a13d6dec82d1, a change was introduced so that duplicate www-authenticate headers weren't ignored.

At the same time, this change unknowingly removed the previous behaviour of ignoring a location header if www_authenticate headers were present (and came first in the list). This resulted in servers returning both location headers and authenticate headers causing a loop in the login process: https://github.com/nextcloud/android/issues/9827

This commit fixes that, while also allowing multiple authentication headers as before.

As a side note, the connection check logic (including this part) is very opaque and should be reworked. `RemoteOperationResult` constructors also show inconsistent behaviour.

Fixes https://github.com/nextcloud/android/issues/9827

### DIFF
--- a/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
+++ b/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
@@ -378,10 +378,10 @@ public class RemoteOperationResult<T extends Object> implements Serializable {
             Header current;
             for (Header httpHeader : httpHeaders) {
                 current = httpHeader;
-                if (HEADER_LOCATION.equals(current.getName().toLowerCase(Locale.US))) {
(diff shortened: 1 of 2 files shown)


AlvaroBrey:
> Can you add a specific test for this? So that when we rework the whole thing, we have more confident that it still works as expected.

There was already a unit test for this case, but it reflected the wrong behaviour (allowing location + www_authenticate at the same time). Fixed in https://github.com/nextcloud/android-library/pull/820/commits/c246898e79ef50406cfb6e2eed06220d1a5aa62a

tobiasKaminsky (approved):

PR #830, opened 2022-02-25
Remove some lombok

tobiasKaminsky:


library/src/main/java/com/owncloud/android/lib/common/Quota.kt
```diff
@@ -45,4 +43,8 @@ data class Quota(
     val quota: Long = 0,
 ) : Parcelable {
     constructor(quota: Long) : this(0, quota = quota)
+
+    override fun toString(): String {
+        return "Quota: $quota, free: $free, used: $used, total: $total, relative: $relative"
+    }
```
AlvaroBrey, line 49: Unneeded unless you want that specific format. `data class`already autogenerates `toString()`

library/src/main/java/com/owncloud/android/lib/resources/activities/model/Activity.kt
```diff
@@ -23,47 +23,43 @@
...
     @SerializedName("object_name")
-    private String objectName;
-    private List<PreviewObject> previews;
+    var objectName: String,
+    var previews: List<PreviewObject>,
+
     @SerializedName("subject_rich")
-    private RichElement richSubjectElement;
-}
+    var richSubjectElement: RichElement
```
AlvaroBrey, line 64: These can all be `val`, since the original class didn't have `@Setter`

AlvaroBrey (changes requested):

AlvaroBrey (approved):

AlvaroBrey (changes requested):
getActivities test is failing

AlvaroBrey:
> getActivities test is failing

Fixed in https://github.com/nextcloud/android-library/pull/830/commits/2861d692427cb7e3abd965c47030747538f95be9

AlvaroBrey (approved):

PR #833, opened 2022-03-08
IPv4FallbackInterceptor: handle IllegalStateException when chain is exhausted

AlvaroBrey:
From play console error:

```
java.lang.IllegalStateException: 
  at okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.kt:99)
  at com.nextcloud.common.IPv4FallbackInterceptor.retryWithIPv4 (IPv4FallbackInterceptor.kt:74)
  at com.nextcloud.common.IPv4FallbackInterceptor.intercept (IPv4FallbackInterceptor.kt:54)
  at okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.kt:109)
```

This is likely caused by https://github.com/square/okhttp/blob/master/okhttp/src/jvmMain/kotlin/okhttp3/internal/http/RealInterceptorChain.kt#L99 (chain can only proceed once)

### DIFF
--- a/src/main/java/com/nextcloud/common/IPv4FallbackInterceptor.kt
+++ b/src/main/java/com/nextcloud/common/IPv4FallbackInterceptor.kt
@@ -71,6 +71,11 @@ class IPv4FallbackInterceptor(private val connectionPool: ConnectionPool) : Inte
         Log_OC.d(TAG, "Error with IPv6, trying IPv4")
         DNSCache.setIPVersionPreference(hostname, true)
         connectionPool.evictAll()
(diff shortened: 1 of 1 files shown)


AlvaroBrey:
Will be useless when okhttp 5.0 is out, as they've implemented Happy Eyeballs: https://github.com/square/okhttp/issues/506#issuecomment-1024256588, https://square.github.io/okhttp/changelogs/changelog/#version-500-alpha4

So let's give up on this for now, and come back when that's released.

PR #839, opened 2022-03-11
Have userId tests also for NextcloudClient

tobiasKaminsky:
I am unsure, if this is the best approach on NextcloudClient.
At least we have then a fitting method name, but on the other hand devs could still use nextcloudClient.userId which is then like plainId, but not known to them.

### DIFF
--- a/src/androidTest/java/com/owncloud/android/lib/common/OwnCloudClientTest.kt
+++ b/src/androidTest/java/com/owncloud/android/lib/common/OwnCloudClientTest.kt
@@ -27,8 +27,10 @@
 package com.owncloud.android.lib.common
 
 import android.net.Uri
+import com.nextcloud.common.NextcloudClient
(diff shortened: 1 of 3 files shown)


AlvaroBrey (commented):
I like the method names for `NextcloudClient`.

As for the code: Rather than duplicating the code and the tests, I'd rather create some public method (`SomethingUtil.encodeUserId`) to encode userIDs that both `NextcloudClient` and `OwncloudClient` can call, and test that instead.

tobiasKaminsky:
> As for the code: Rather than duplicating the code and the tests, I'd rather create some public method (`SomethingUtil.encodeUserId`) to encode userIDs that both `NextcloudClient` and `OwncloudClient` can call, and test that instead.

My aim is that we can get rid of OwnCloudClient very soon, so this might be unnecessary work?

AlvaroBrey (approved):

PR #844, opened 2022-03-24
Parse more attributes for shares

AlvaroBrey:


library/src/main/java/com/owncloud/android/lib/resources/shares/OCShare.java
```diff
@@ -222,6 +229,9 @@ public void readFromParcel(Parcel source) {
         isPasswordProtected = source.readInt() == 1;
         hideFileDownload = source.readInt() == 1;
         label = source.readString();
+        hasPreview = source.readInt() == 1;
+        mimetype = source.readString();
+        ownerDisplayName = source.readString();
```
tobiasKaminsky, line 234: Can you create a test for this?
We have already a test here: https://github.com/nextcloud/android-library/blob/master/library/src/androidTest/java/com/owncloud/android/lib/resources/shares/GetSharesRemoteOperationIT.java#L126

library/src/main/java/com/owncloud/android/lib/resources/files/SearchRemoteOperation.java
```diff
@@ -54,7 +54,6 @@
         FAVORITE_SEARCH, // get all favorite files/folder
         RECENTLY_MODIFIED_SEARCH, // get files/folders that were modified within last 7 days, ordered descending by time
         PHOTO_SEARCH, // gets all files with mimetype "image/%"
-        SHARED_SEARCH, // show all shares
```
tobiasKaminsky, line 57: Is this really not needed anylonger?
Will lead to crash here: https://github.com/nextcloud/android/blob/23a72722d340060c797881d56281baf74a4c26c2/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java#L1701

Or is there a corresponding PR to remove it? :)

AlvaroBrey: Removed in https://github.com/nextcloud/android/pull/10015/files#diff-814a2c94173812cefd41e8a282e724941bbee46ae5c3332602936360609fe578L1701

It's only ever used in that `if`, which means it doesn't really do anything.

tobiasKaminsky: I was sure that you already did this :) 
We then just need to make sure to merge #10015 before this one.

library/src/main/java/com/owncloud/android/lib/resources/shares/OCShare.java
AlvaroBrey: What would you like me to test? The test you linked only checks share type and path.

library/src/main/java/com/owncloud/android/lib/resources/files/SearchRemoteOperation.java
AlvaroBrey: > I was sure that you already did this :) We then just need to make sure to merge #10015 before this one.

That is a bit of a problem because #10015 requires the new attributes in this PR lol. I can remove this change and we'll clean it up at a later date.

tobiasKaminsky: We have this from time to time. Then we usually just need to merge it within a very short time gap.
As it is not a mandatory change, it might be easier for this time to remove it.

library/src/main/java/com/owncloud/android/lib/resources/shares/OCShare.java
tobiasKaminsky: Correct, there you can add mimetime, hasPreview and ownerDisplayName check.
E.g. in this case it is a folder, with no preview and owner is the user "admin" (I think).
Basically just to check that those attributes are really returned by server and parsed correctly.

library/src/main/java/com/owncloud/android/lib/resources/files/SearchRemoteOperation.java
AlvaroBrey: Reverted in https://github.com/nextcloud/android-library/pull/844/commits/e184d27168a58b9dbdb3dc6226ea5c4fb44e1acb

library/src/main/java/com/owncloud/android/lib/resources/shares/OCShare.java
AlvaroBrey: Done https://github.com/nextcloud/android-library/pull/844/commits/e2baf8d33f5ef6baf212ee75a31e6dfef24bdf00

tobiasKaminsky (approved):

PR #854, opened 2022-04-07
Encode single quotes in usernames

AlvaroBrey:
Fixes https://github.com/nextcloud/android/issues/10053


tobiasKaminsky (approved):

tobiasKaminsky:
userId should never been anything else than a-z0-9, but now it is too late… ;)

Issue #868, opened 2022-04-20
Get access to Etag after at the end of an UploadFileRemoteOperation


vince-bourgmayer:
Hello,

First of all, thanks for your work and your Library.
I'm looking for a way to read latest etag of a file after it has been uploaded with UploadFileRemoteOperation.

This could easily been read from HTTP's result header 
in https://github.com/nextcloud/android-library/blob/f0b28e537afcc113473783192c60fd13c573f30b/library/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java#L218

and 

https://github.com/nextcloud/android-library/blob/f0b28e537afcc113473783192c60fd13c573f30b/library/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java#L409

There is different way to do it but I guess the easiest one is to store those headers in a class field and provide a getHeaders() to access it.  The main issue I guess is to provide something generic to match with Apache Commons & OkHttp while you're migrating :-/

Edit:
Here is my proposal:
https://github.com/nextcloud/android-library/commit/415d824694a155eca4c3b724c16603002b0a316f

TheScarastic:
@AlvaroBrey any insights if patch looks fine to you we could make a pull request

vince-bourgmayer:
Pull request created:
https://github.com/nextcloud/android-library/pull/869

AlvaroBrey:
> I'm looking for a way to read latest etag of a file after it has been uploaded with UploadFileRemoteOperation.

The correct way to do this is to have `UploadFileRemoteOperation`  save the required data  (in this case the etag) in `RemoteOperationResult.resultData`, and grab that `resultData` when needed.

Technically, `RemoteOperationResult` should abstract the details of HTTP requests and results from its clients, even though it does already expose `httpStatus` and  some location stuff. 

I can't really come up with use cases which could be solved by exposing the headers, but not by adding result data to the relevant operation, so my initial assesment of this is that it's not worth polluting the class.

vince-bourgmayer:
Hi @AlvaroBrey,

Thanks for your feedback.
The way you suggest is the way I first thought to do it;

I also tried to extends the UploadFileRemoteOperation in my application with the will to just override
`protected RemoteOperationResult uploadFile(OwnCloudClient client) throws IOException` which is the only part I saw where I could read the etag.

But then I couldn't access mPutMethod because I'm not in the same package.
[...]

PR #869, opened 2022-04-20
 Add a getHeaders() to get headers from result.

vince-bourgmayer:
- Add a class field 'mHeaders' as private Headers (from okHttp3)
- Add a getter for 'mHeaders' field
- Update constructor wich receive headers:
    1. Convert old Apache commons headers into okHttp3 header then add them into new mHeaders class field.
    2. set OkHttp3 headers as value for new mHeaders class field


AlvaroBrey:
See discussion in #868 before reviewing

PR #870, opened 2022-04-21
Update UploadFileRemoteOperation.java to get access to etag from resp…

vince-bourgmayer:
…onse header

- Add a private static final value which is the response header's name for etag.
- update 'RemoteOperationResult uploadFile(OwnCloudClient client)' to check for this header, and set value as RemoteOperationResult's data.

### DIFF
--- a/library/src/main/java/com/owncloud/android/lib/resources/files/UploadFileRemoteOperation.java
+++ b/library/src/main/java/com/owncloud/android/lib/resources/files/UploadFileRemoteOperation.java
@@ -35,6 +35,7 @@ import com.owncloud.android.lib.common.operations.RemoteOperation;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 
 import org.apache.commons.httpclient.DefaultHttpMethodRetryHandler;
+import org.apache.commons.httpclient.Header;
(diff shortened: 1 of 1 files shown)


AlvaroBrey (changes requested):
Looks fine, but since the `RemoteOperationResult` has result data, it should be parameterized.

For this:
- `UploadFileRemoteOperation` should extend `RemoteOperation<String>` rather than the raw type
- Variables and return types should be `RemoteOperationResult<String>` (or `RemoteOperationResult<>` where applicable) rather than the raw type.

AlvaroBrey (approved):

AlvaroBrey:
@vince-bourgmayer Please fix DCO (https://github.com/nextcloud/android-library/pull/870/checks?check_run_id=6127967707) and this will be ready to merge

vince-bourgmayer:
> @vince-bourgmayer Please fix DCO (https://github.com/nextcloud/android-library/pull/870/checks?check_run_id=6127967707) and this will be ready to merge

@AlvaroBrey DCO has been fixed

AlvaroBrey (approved):

AlvaroBrey:
Thank you @vince-bourgmayer !

PR #872, opened 2022-04-22
IPv4FallbackInterceptor: don't reuse same call when retrying

AlvaroBrey:
Successor to #833

Signed-off-by: Álvaro Brey Vilas <alvaro.brey@nextcloud.com>


AlvaroBrey:
Superseded by #897

PR #881, opened 2022-05-02
Switching to dav4jvm

tobiasKaminsky:
Heavily under construction 🚧
If you have any feedback, please comment :)


tobiasKaminsky:
@AlvaroBrey if you have a bit of time, any feedback would be great 💙

library/src/main/java/com/nextcloud/operations/dav/OkHttpDavMethodBase.kt
```diff
@@ -0,0 +1,55 @@
...
+
+abstract class OkHttpDavMethodBase {
+    var response: Response? = null
+
+    fun execute(nextcloudClient: NextcloudClient) {
+        nextcloudClient.client = nextcloudClient.client
+            .newBuilder()
+            .followRedirects(false)
+            .addInterceptor(BasicAuthInterceptor(nextcloudClient.credentials))
+            .build()
```
AlvaroBrey, line 37: Can you explain the context of this a bit? Why do we need a basic auth interceptor only for dav calls, and why can't it follow redirects?

Additionally this would create a new client for every call which is going to be very inefficient, AND it also modifies the client for future users of `NextcloudClient`, which we don't want.

library/src/main/java/com/nextcloud/common/NextcloudClient.kt
```diff
@@ -118,6 +119,11 @@ class NextcloudClient private constructor(
         return method.execute(this)
     }
 
+    @Throws(Exception::class)
+    fun execute(method: OkHttpDavMethodBase) {
```
AlvaroBrey, line 123: Ideally we'd have the dav method extend from `OkHttpMethodBase` as well, and `NextcloudClient` would not care about the kind of method it's implementing.

I guess this is due to the different execution model of `dav4jvm` (which doesn't let you access the methods it creates, and instead executes them directly).

If this is not possible, then we should have `OkHttpMethodBase` and `OkHttpDavMethodBase` at least implement a common interface (like `NextcloudClientMethod` or something) and have a single `execute` method that takes that interface.

library/src/main/java/com/nextcloud/operations/dav/OkHttpDavMethodBase.kt
AlvaroBrey: I see that this is because we can't manipulate anything from the call that `dav4jvm` does. In this case, I'd rather use the `BasicAuthInterceptor` for all calls, including the non-Dav ones, rather than handling authorization in `OkHttpMethodBase`.

library/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
```diff
@@ -316,6 +320,10 @@ public RemoteOperationResult(boolean success, OkHttpMethodBase httpMethod) {
         this(success, httpMethod.getStatusCode(), httpMethod.getStatusText(), httpMethod.getResponseHeaders());
     }
 
+    public RemoteOperationResult(boolean success, OkHttpDavMethodBase davMethod) {
+        this(success, davMethod.statusCode());
+    }
```
AlvaroBrey, line 325: Again, it would be better to have `OkHttpMethodBase` and `OkHttpDavMethodBase` share a hierarchy or implement a common interface, so that `RemoteOperationResult` wouldn't need to care which of them it's dealing with.

library/src/main/java/com/owncloud/android/lib/common/utils/WebDavFileUtils.java
```diff
@@ -72,11 +90,92 @@ public ArrayList<RemoteFile> readData(MultiStatus remoteData,
...
+
+        return list;
+    }
+
+    public RemoteFile parseResponse(Response response, Uri filesDavUri) {
+        RemoteFile remoteFile = new RemoteFile();
+        String path = response.getHref().toString().split(filesDavUri.getEncodedPath(), 2)[1].replace("//", "/");
+
+        for (Property property : response.getProperties()) {
+            if (property instanceof NCEtag) {
```
AlvaroBrey, line 116: I don't like all these class checks, but I guess that's just how `dav4jvm` works. I'd suggest at least converting this class to Kotlin, so you could at least use a single `when(property) { is NCETag -> ...`  and avoid the casts

library/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperation.java
```diff
@@ -448,11 +448,14 @@ public void run() {
...
-     * 
-     * @return      Current client instance to access the remote server.
+     *
+     * @return Current client instance to access the remote server.
      */
     public final OwnCloudClient getClient() {
         return mClient;
     }
 
+    public final NextcloudClient getClientNew() {
```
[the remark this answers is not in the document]
AlvaroBrey: In more general terms, I don't think the `RemoteOperation` should expose the client at all; if an operation needs an userId to work, it should be part of its parameters directly.

library/src/main/java/com/owncloud/android/lib/resources/files/SearchRemoteOperation.java
```diff
@@ -169,5 +175,82 @@ protected RemoteOperationResult<List<RemoteFile>> run(OwnCloudClient client) {
...
+                    capability,
+                    startDate,
+                    endDate);
+
+            // disable redirect
+            OkHttpClient disabledRedirectClient = client.getClient()
+                    .newBuilder()
+                    .followRedirects(false)
+                    .authenticator(new NextcloudAuthenticator(client.getCredentials(), "Authorization"))
+                    .build();
```
AlvaroBrey, line 207: I don't think it's a good idea to create a new OkHttpClient for every request...  the `NextcloudClient` passed to `run` should already be authenticated IMO. Not sure about the redirects or why they are disabled.

library/src/main/java/com/owncloud/android/lib/resources/files/SearchRemoteOperation.java
```diff
@@ -169,5 +175,82 @@ protected RemoteOperationResult<List<RemoteFile>> run(OwnCloudClient client) {
...
+            // Add data to the result
+            if (result.isSuccess()) {
+                result.setResultData(list);
+            }
+        } catch (Exception e) {
+            result = new RemoteOperationResult<>(e);
+        } finally {
+            if (searchMethod != null) {
+                searchMethod.releaseConnection();  // let the connection available for other methods
+            }
```
AlvaroBrey, line 243: This does not seem necessary, as the search method is never called (only used to get the query string)

library/src/main/java/com/owncloud/android/lib/resources/files/ToggleFavoriteRemoteOperation.java
```diff
@@ -89,12 +101,49 @@ protected RemoteOperationResult run(OwnCloudClient client) {
...
+
+        List<Property.Name> removeProperties = new ArrayList<>();
+        Map<Property.Name, String> newProperties = new HashMap<>();
+
+        // disable redirect
+        OkHttpClient disabledRedirectClient = client.getClient()
+                .newBuilder()
+                .followRedirects(false)
+                .authenticator(new NextcloudAuthenticator(client.getCredentials(), "Authorization"))
+                .build();
```
AlvaroBrey, line 125: Same comments about creating a new client for every request apply.

library/src/main/java/com/owncloud/android/lib/resources/files/ToggleFavoriteRemoteOperation.java
```diff
@@ -89,12 +101,49 @@ protected RemoteOperationResult run(OwnCloudClient client) {
...
+        String encodedPath = (client.getUserId() + Uri.encode(filePath)).replace("%2F", "/");
+        String fullFilePath = webDavUrl + "/files/" + encodedPath;
+
+        boolean resultCode = false;
+
+        new DavResource(disabledRedirectClient,
+                HttpUrl.get(fullFilePath))
+                .proppatch(newProperties, removeProperties, (response, hrefRelation) -> {
+                    //resultCode = response.isSuccess();
+                });
```
AlvaroBrey, line 143: We should likely wrap `DavResource` somehow, I don't like how coupled it's becoming to these operations.

library/src/main/java/com/owncloud/android/lib/resources/status/OCCapability.kt
```diff
@@ -27,14 +27,15 @@ package com.owncloud.android.lib.resources.status
 /**
  * Contains data of the Capabilities for an account, from the Capabilities API
  */
-class OCCapability {
+class OCCapability(
```
AlvaroBrey, line 30: Why this change? If it's just for the tests, use `apply` instead. I don't see a reason to have only these three specific properties as constructor parameters.

AlvaroBrey (commented):
- [ ] Unused submodule?

PR #886, opened 2022-05-04
OCCapability: restore nullability of optional strings

AlvaroBrey:
Fixes https://github.com/nextcloud/android/issues/10162


tobiasKaminsky (dismissed):

tobiasKaminsky:
Can you add some sort of test for this?

AlvaroBrey:
> Can you add some sort of test for this?

Added a "unit test", which will actually fail to compile if we make those fields non-nullable. That's the best I can do.

tobiasKaminsky (approved):

PR #897, opened 2022-05-13
Bump okhttp to 5.0.0-alpha7 for Happy Eyeballs support

AlvaroBrey:
This should solve everything related to IPv4/6 in OkHTTP. Once we get everything ported to NextcloudClient, we can also remove DNSCache entirely.

From [okhttp's CHANGELOG](https://github.com/square/okhttp/blob/master/CHANGELOG.md):

> The alpha releases in the 5.0.0 series have production-quality code and an unstable API. We expect to make changes to the APIs introduced in 5.0.0-alpha.X. These releases are safe for production use and 'alpha' strictly signals that we're still experimenting with some new APIs. If you're eager for the fixes or features below, please upgrade.

So it looks like it's safe to upgrade even if it says `alpha` in the name, and we kinda need this feature.

Fixes https://github.com/nextcloud/android/issues/10039
Fixes https://github.com/nextcloud/android/issues/10238


tobiasKaminsky (approved):

tobiasKaminsky:
Cool. Let's try it 🤞

PR #932, opened 2022-07-04
ktlint plugin switch

### LINKED ISSUE #929: ktlint not working for sources outside the default package
This is clearly visible by using the ktlint plugin in Android Studio: Files outside com.nextcloud.android.lib do not get reported in ./gradlew ktlintCheck, but do get reported in the IDE.

AlvaroBrey:
- build: Replace ktlint plugin with spotless
- Fix pre-commit hook to use spotless
- Fix git hooks copy logic
- Update check workflow to use spotless
- Auto-fix ktlint format issues

Fixes #929

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,7 +12,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                task: [ detekt, ktlintCheck, lint ]
+                task: [ detekt, spotlessKotlinCheck, lint ]
         steps:
             -   uses: actions/checkout@v3
             -   name: Set up JDK 11
--- a/build.gradle
+++ b/build.gradle
@@ -16,3 +16,19 @@ subprojects {
 task clean(type: Delete) {
     delete rootProject.buildDir
 }
+
+task installGitHooks(type: Copy, group: "development") {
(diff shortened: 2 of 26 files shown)


AlvaroBrey:
Note to self: try Spotless instead. Seems to be widely used and may not need gradle 7.3

AlvaroBrey:
Spotless also can handle formatting for Markdown, Java, Groovy, XML,  etc. We can later think about moving this to android-config and using it in all repos instead of individual formatting configurations

tobiasKaminsky (dismissed):

tobiasKaminsky:
@AlvaroBrey do you have an idea why codecov is so low? 36% compared to 48%?

AlvaroBrey:
> @AlvaroBrey do you have an idea why codecov is so low? 36% compared to 48%?

Coverage is still 48.73% (https://app.codecov.io/gh/nextcloud/android-library/compare/932?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr%20comments&utm_term=nextcloud)

36% is the coverage of the _diff_ i.e. only the changes in this PR. Looks like we have it configured to fail if the diff coverage is lower than the project coverage, which makes sense.

PR #954, opened 2022-08-23
Move assemble check to Github actions

AlvaroBrey:
Less likely to break and easier to restart

### DIFF
--- a/.github/workflows/assemble.yml
+++ b/.github/workflows/assemble.yml
@@ -1,0 +1,30 @@
+name: Assemble
+
+on:
+    pull_request:
+        branches: [ master, stable-* ]
+    push:
+        branches: [ master, stable-* ]
+
+jobs:
+    assemble:
+        runs-on: ubuntu-latest
+        steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/assemble.yml
```diff
@@ -0,0 +1,30 @@
...
+                    java-version: 11
+            -   name: Setup JVM options
+                run: |
+                    mkdir -p "$HOME/.gradle"
+                    echo "org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError" > "$HOME/.gradle/gradle.properties"
+            -   name: Assemble
+                uses: gradle/gradle-build-action@v2
+                with:
+                    arguments: assemble
+            -   uses: actions/upload-artifact@v3
```
tobiasKaminsky, line 27: Do we need to upload it?

AlvaroBrey: removed

PR #985, opened 2022-11-08
ChunkedFileRemoteOperation: correctly use lastModificationTimestamp

AlvaroBrey:
Fixes https://github.com/nextcloud/android/issues/10900


AlvaroBrey:
> Fixes [nextcloud/android#10900](https://github.com/nextcloud/android/issues/10900)

Confirmed by affected users

tobiasKaminsky:
Right. Seems we did converting twice.
Can you add info that lastModificationTimestamp must be in seconds, according to unix time?

AlvaroBrey:
> Right. Seems we did converting twice. 

No, the problem here was that `ChunkedFileRemoteOperation` was ignoring the  passed in `lastModificationTimestamp` and instead using `file.lastModified()`, which gives the value of the copy, not the original.

> Can you add info that lastModificationTimestamp must be in seconds, according to unix time?

Done.

tobiasKaminsky (approved):

PR #1005, opened 2022-11-26
Support client certificates

stephanritscher:
This change adds support for configurations where the webserver requests a client certificate, e.g. via nginx configuration options ssl_client_certificate and ssl_verify_client. The client certificate handling is done via InteractiveKeyManager which prompts for a client certificate from a file or the devices keystore.

This change is NOT about client certificate authentication to the nextcloud server instance. The regular authentication mechanisms will be used as soon as the communication on TLS level is established.

The change addresses ticket #603, while not being a generic solution IMO.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -48,6 +48,7 @@ configurations {
 }
 
 dependencies {
+    implementation 'com.github.stephanritscher:InteractiveKeyManager:0.1'
     implementation 'org.apache.jackrabbit:jackrabbit-webdav:2.13.5'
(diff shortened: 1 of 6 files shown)


Torqu3Wr3nch:
> The change addresses ticket #603, while not being a generic solution IMO.

I think @stephanritscher probably means https://github.com/nextcloud/android/issues/603. Thanks again for doing this, Stephan.

AlvaroBrey:
app PR: https://github.com/nextcloud/android/pull/11099

AlvaroBrey:
Closed per https://github.com/nextcloud/android/pull/11099#issuecomment-1335052154

PR #1007, opened 2022-12-09
Build unit and androidTest code before committing

tobiasKaminsky:
Sometimes it happens that a code change kills tests.
This prevents us time-intensive roundtrips via CI.


AlvaroBrey:
Since we discarded this on Files, should we close it here too?

PR #1009, opened 2022-12-12
use qualified assert import

tobiasKaminsky:


AlvaroBrey (dismissed):
I don't have a preference one way or the other, so :+1:

AlvaroBrey (approved):

PR #1027, opened 2023-01-05
fix findings of detekt

tobiasKaminsky:
Detekt is now at 0 :muscle: 

### DIFF
--- a/library/src/main/java/com/nextcloud/common/NextcloudClient.kt
+++ b/library/src/main/java/com/nextcloud/common/NextcloudClient.kt
@@ -124,6 +124,7 @@ class NextcloudClient private constructor(
     }
 
     @Throws(IOException::class)
+    @Suppress("NestedBlockDepth")
     fun followRedirection(method: OkHttpMethodBase): RedirectionPath {
         var redirectionsCount = 0
         var status = method.getStatusCode()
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/nextcloud/common/NextcloudClient.kt
```diff
@@ -146,30 +146,26 @@ class NextcloudClient private constructor(
                 method.uri = location
                 var destination = method.getRequestHeader("Destination")
```
AlvaroBrey, line 148: You've removed the logic to check `destination` if `Destination` is not present. Is this intentional? Shouldn't we instead check for `destination.isEmpty()`?

tobiasKaminsky: Android studio said that this is never empty.
I assumed that this would be a function by OkHhttp, but it is ours:
https://github.com/nextcloud/android-library/blob/a33586c0e68287ce428cf0358ad3e1df4d86ae18/library/src/main/java/com/nextcloud/common/OkHttpMethodBase.kt#L127

And thus you are right, it would kill this check.
I'll revert it.

AlvaroBrey (approved):

PR #1039, opened 2023-01-17
Increase wait time for `wait_for_server`

AlvaroBrey:
This script waits for the server to wake up during CI before launching tests.
In this library it only waited for 100 seconds instead of the 500 seconds it waits in the Files app.

This caused "failed to wait for server" in some occasions where the server image needs to be downloaded
and there's slow internet or poor performance.

### DIFF
--- a/scripts/wait_for_server.sh
+++ b/scripts/wait_for_server.sh
@@ -10,7 +10,7 @@ until [[ $status = "false" ]]; do
 
     if [[ "$status" =~ "false" || "$status" = "" ]]; then
         let "counter += 1"
-         if [[ $counter -gt 10 ]]; then
+         if [[ $counter -gt 50 ]]; then
             echo "Failed to wait for server"
             exit 1
         fi


scripts/wait_for_server.sh
```diff
@@ -1,20 +1,23 @@
 #!/usr/bin/env bash
 
-counter=0
 status=""
+counter=0
+checkcounter=0
```
tobiasKaminsky, line 5: Just a small nitpick, but why do we need two counters?

AlvaroBrey: It's this way on Files :shrug:  I can remove it from both if needed

AlvaroBrey: All changes except wait time increase are now undone

tobiasKaminsky (approved):

scripts/wait_for_server.sh
tobiasKaminsky: Hm. Strange, I have no idea why this is needed.
At some point we shall move this also to common…

PR #1068, opened 2023-02-14
feat(download): Use specific code and exception when download fails to create local file

AlvaroBrey:


AlvaroBrey:
Superseded by #1069 (as the previous branchname was not buildable as snapshot)

PR #1069, opened 2023-02-14
feat(download): Use specific code and exception when download fails to create local file

AlvaroBrey:


tobiasKaminsky:
This needs a corresponding files PR, otherwise new ResultCode is not correct understood and "unknown error" is shown, right?

AlvaroBrey:
> This needs a corresponding files PR, otherwise new ResultCode is not correct understood and "unknown error" is shown, right?

Yes, PR is https://github.com/nextcloud/android/pull/11384

PR #1079, opened 2023-03-08
fix: FileVersion: remove duplicate getters

AlvaroBrey:
These getters are generated by Lombok; so it's failing to compile. Test before/after with
gw clean :library:compileDebugJavaWithJavac --no-build-cache

Introduced by #651, not sure how we missed it there


tobiasKaminsky:
As we want to remove lombok, I think it is better to replace this, with either data class, or with plain Java right now.
Especial this 0 for getFileLength() would otherwise be lost.

tobiasKaminsky (approved):

PR #1114, opened 2023-04-20
Use resultData

tobiasKaminsky:


AndyScherzinger:
@tobiasKaminsky your newly added test fails 😢 https://github.com/nextcloud/android-library/pull/1114#issuecomment-1566223081

PR #1215, opened 2023-10-10
Handle user disabled exception as its own result code

tobiasKaminsky:


ZetaTom (dismissed):

alperozturk96 (approved):

alperozturk96 (dismissed):

alperozturk96 (approved):

ZetaTom (approved):

alperozturk96 (approved):

library/src/main/java/com/owncloud/android/lib/common/operations/ExceptionParser.java
```diff
@@ -46,6 +46,8 @@ public class ExceptionParser {
     private static final String INVALID_PATH_EXCEPTION_STRING = "OC\\Connector\\Sabre\\Exception\\InvalidPath";
     private static final String INVALID_PATH_EXCEPTION_UPLOAD_STRING = "OCP\\Files\\InvalidPathException";
     private static final String VIRUS_EXCEPTION_STRING = "OCA\\DAV\\Connector\\Sabre\\Exception\\UnsupportedMediaType";
+    private static final String USER_DISABLED_MESSAGE = "OC\\User\\LoginException: User disabled";
+    private static final String SERVICE_UNAIVALABLE_STRING = "Sabre\\DAV\\Exception\\ServiceUnavailable";
```
joshtrichards, line 50:
```suggestion
    private static final String SERVICE_UNAVAILABLE_STRING = "Sabre\\DAV\\Exception\\ServiceUnavailable";
```

library/src/main/java/com/owncloud/android/lib/common/operations/ExceptionParser.java
```diff
@@ -92,6 +94,10 @@ public boolean isVirusException() {
         return VIRUS_EXCEPTION_STRING.equalsIgnoreCase(exception) && message.startsWith("Virus");
     }
 
+    public boolean isUserDisabledException() {
+        return SERVICE_UNAIVALABLE_STRING.equals(exception) && message.equalsIgnoreCase(USER_DISABLED_MESSAGE);
```
joshtrichards, line 98:
```suggestion
        return SERVICE_UNAVAILABLE_STRING.equals(exception) && message.equalsIgnoreCase(USER_DISABLED_MESSAGE);
```

library/src/main/java/com/owncloud/android/lib/common/operations/ExceptionParser.java
```diff
@@ -30,6 +30,8 @@ public class ExceptionParser {
     private static final String INVALID_PATH_EXCEPTION_STRING = "OC\\Connector\\Sabre\\Exception\\InvalidPath";
     private static final String INVALID_PATH_EXCEPTION_UPLOAD_STRING = "OCP\\Files\\InvalidPathException";
     private static final String VIRUS_EXCEPTION_STRING = "OCA\\DAV\\Connector\\Sabre\\Exception\\UnsupportedMediaType";
+    private static final String USER_DISABLED_MESSAGE = "OC\\User\\LoginException: Account disabled";
```
joshtrichards, line 33: I don't think we can include the 'Account disabled` part since it gets translated on the backend:

https://github.com/nextcloud/server/blob/57c71f8b9b34eb2af2914d11a9a6993ddf8deedf/lib/private/legacy/OC_User.php#L159

https://github.com/nextcloud/server/blob/57c71f8b9b34eb2af2914d11a9a6993ddf8deedf/lib/private/User/Session.php#L321

joshtrichards (commented):
If we make this more generic (for all LoginException's) we could also maybe cover other scenarios we're apprently not handling well like:

`"message": "OC\\User\\LoginException: Password is expired, please use forgot password method to reset"` (#8697)

PR #1242, opened 2023-11-16
Update Old Metadata API

alperozturk96:


library/src/main/java/com/owncloud/android/lib/common/network/WebdavEntry.kt
```diff
@@ -577,8 +577,8 @@ class WebdavEntry constructor(ms: MultiStatusResponse, splitElement: String) {
         const val EXTENDED_PROPERTY_LOCK_TIMEOUT = "lock-timeout"
         const val EXTENDED_PROPERTY_LOCK_TOKEN = "lock-token"
         const val EXTENDED_PROPERTY_SYSTEM_TAGS = "system-tags"
-        const val EXTENDED_PROPERTY_METADATA_SIZE = "file-metadata-size"
-        const val EXTENDED_PROPERTY_METADATA_GPS = "file-metadata-gps"
+        const val EXTENDED_PROPERTY_METADATA_PHOTOS_SIZE = "file-metadata-photos-size"
+        const val EXTENDED_PROPERTY_METADATA_PHOTOS_GPS = "file-metadata-photos-gps"
```
artonge, line 581:
```suggestion
        const val EXTENDED_PROPERTY_METADATA_PHOTOS_SIZE = "files-metadata-photos-size"
        const val EXTENDED_PROPERTY_METADATA_PHOTOS_GPS = "files-metadata-photos-gps"
```

artonge (dismissed):
The metadata names looks good. Can't judge the rest of the code.

tobiasKaminsky:
The tests fail.
This then will also fail on e.g. stable 27.
So we need to have both metadata names, as per https://github.com/nextcloud/files-clients/issues/48#issuecomment-1814056368

Both can set the same property, as only either new or old can be there.

AndyScherzinger:
This is where it gets a bit tricky. The client supports all servers v16+, so the client needs to be able to read both value not just v28.
With this PR you don't have backwards compatibility for 27/Hub6.

cc @tobiasKaminsky

tobiasKaminsky:
> This is where it gets a bit tricky. The client supports all servers v16+, so the client needs to be able to read both value not just v28.

Please read what I posted above, as I already explained it, together with a solutionn.

tobiasKaminsky:
We now test against stable 27, so we can test that it still works for 27 and for master / 28

AndyScherzinger:
> Please read what I posted above, as I already explained it, together with a solutionn.

whoops, I seems to have missed that post, even though it was literally the last one before I posted my reply 👍

tobiasKaminsky:
> Values comes as XML no need to use Gson anymore but for v27 still Gson must be used I guess.

@artonge why do we we have breaking changes on an existing api? 
@alperozturk96 if new one is XML and old one is JSON, you can handle this different based on old/new property name, I guess. So it parses it differently, but in RemoteFile the same value is being stored.

artonge:
> why do we we have breaking changes on an existing api?

The old API was private on purpose, it was unfinished work.

tobiasKaminsky:
```
11-29 13:17:34.073  8199  8223 E ReadFolderRemoteOperation: org.xmlpull.v1.XmlPullParserException: Unexpected token (position:TEXT [org.apache.harm...@1:97 in java.io.StringReader@d491920)
11-29 13:17:34.073  8199  8223 E ReadFolderRemoteOperation: 	at org.kxml2.io.KXmlParser.next(KXmlParser.java:432)
11-29 13:17:34.073  8199  8223 E ReadFolderRemoteOperation: 	at org.kxml2.io.KXmlParser.next(KXmlParser.java:313)
11-29 13:17:34.073  8199  8223 E ReadFolderRemoteOperation: 	at com.owncloud.android.lib.resources.files.model.ImageDimensionKt.parseImageDimensions(ImageDimension.kt:51)
11-29 13:17:34.073  8199  8223 E ReadFolderRemoteOperation: 	at com.owncloud.android.lib.common.network.WebdavEntry.<init>(WebdavEntry.kt:429)
```

Now it fails for me, locally with this PR, with above error.
Photos app needs to be installed to have generated metadata.

library/src/androidTest/java/com/owncloud/android/lib/resources/files/ReadFileRemoteOperationIT.kt
```diff
@@ -82,7 +82,7 @@ class ReadFileRemoteOperationIT : AbstractIT() {
         testOnlyOnServer(NextcloudVersion.nextcloud_27)
 
         @Suppress("Detekt.MagicNumber")
-        assertEquals(GeoLocation(49.99679166666667, 8.67198611111111), remoteFile.geoLocation)
+        assertEquals(GeoLocation(49.996791666667, 8.6719861111111), remoteFile.geoLocation)
```
alperozturk96, line 85: Test fixed with correct values.

tobiasKaminsky:
> master-IT test failed: https://www.kaminsky.me/nc-dev/library-integrationTests/2327-IT-master-06-32/debug/

Without photos app it only shows one failing test: metadata, which is expected.

tobiasKaminsky (approved):

PR #1328, opened 2024-01-25
Update WebdavEntry.kt to handle images that are partially missing GPS…

mquin2003:
… metadata

This change to WebdavEntry.kt prevents a bug in the Android app. 

The bug occurs when the client loads a directory that contains an image that is partially (not completely) missing some GPS exif data. When the client attempts to load this directory, the app displays a "Server not available" message and fails to refresh.

More details are available about this bug in Issue [#12379](https://github.com/nextcloud/android/issues/12379) of the nextcloud/android repository. 

I also apologize in advance if there's anything wrong with this PR as it's my first time doing one. Thank you! :)


alperozturk96:
@mquin2003 Same thing can happen for WebDavEntry#L431. (_EXTENDED_PROPERTY_METADATA_PHOTOS_SIZE_) this one also new property. Could you also try catch that section?

I didn't understand one part if new prop was not empty (_EXTENDED_PROPERTY_METADATA_PHOTOS_GPS_) how come **org.apache.harmony.xml.dom.ElementImpl cannot be cast to java.util.ArrayList**. Backend should return same type if we have new prop, right? @tobiasKaminsky @artonge 

I am expecting from backend array of some element. Then iterating over all array get values with given tagName. During our tests it worked and data was there always correctly.

Screenshot 2024-02-05 at 12 33 11

If backend may return multiple different type when new property already there we should cast that new type rather then use try catch.

PR #1336, opened 2024-01-28
Add dependency checksums and signing key information

AndyScherzinger:


AndyScherzinger:
@tobiasKaminsky any idea or help on why https://github.com/nextcloud/android-library/actions/runs/7686533038/job/20945497824?pr=1336 might fail? The log doesn't give me any pointers 😢

PR #1363, opened 2024-02-26
Assistant Endpoints

alperozturk96:


library/src/main/java/com/owncloud/android/lib/resources/assistant/CreateTaskRemoteOperation.kt
```diff
@@ -0,0 +1,60 @@
+/*
```
tobiasKaminsky, line 1: Can you change license to match the other ones inside this library? MIT

tobiasKaminsky (approved):

PR #1471, opened 2024-06-12
Make run() always work, even with old ocClient

tobiasKaminsky:
This reduces risk of breaking our client.


ZetaTom (commented):
In my tests this has worked perfectly so far.

PR #1472, opened 2024-06-12
migrate CheckRemoteWipeRemoteOperation to NextcloudClient

tobiasKaminsky:


library/src/androidTest/java/com/owncloud/android/AbstractIT.java
```diff
@@ -89,7 +90,7 @@ public static void beforeAll() throws InterruptedException,
 
         url = Uri.parse(arguments.getString("TEST_SERVER_URL"));
         String loginName = arguments.getString("TEST_SERVER_USERNAME");
-        String password = arguments.getString("TEST_SERVER_PASSWORD");
+        password = arguments.getString("TEST_SERVER_PASSWORD");
```
ZetaTom, line 93: I'm not sure why this variable was made static.

library/src/main/java/com/owncloud/android/lib/resources/users/CheckRemoteWipeRemoteOperation.java
```diff
@@ -7,65 +7,78 @@
...
                 String response = postMethod.getResponseBodyAsString();
 
                 JSONObject json = new JSONObject(response);
 
                 if (json.getBoolean(WIPE)) {
-                    result = new RemoteOperationResult(true, postMethod);
+                    result = new RemoteOperationResult<>(true, postMethod);
                 } else {
-                    result = new RemoteOperationResult(false, postMethod);
+                    result = new RemoteOperationResult<>(false, postMethod);
```
ZetaTom, line 72: If I'm not mistaken, this if statement can be simplified to the following expression:

```java
result = new RemoteOperationResult<>(json.getBoolean(WIPE), postMethod);
```

PR #1485, opened 2024-06-25
base url shall never be null

tobiasKaminsky:


ZetaTom (approved):
Everything works after rebasing onto master.

PR #1578, opened 2024-11-14
add transifex support

tobiasKaminsky:


.tx/config.license
```diff
@@ -0,0 +1,2 @@
+SPDX-FileCopyrightText: 2018-2023 Nextcloud GmbH and Nextcloud contributors
+SPDX-License-Identifier: GPL-3.0-or-later
```
AndyScherzinger, line 2: GPL or MIT, the later would match the libs license

AndyScherzinger (approved):

.tx/config.license
tobiasKaminsky: Ah, right. I c&p from SSO.

nickvergessen:
Oh, I missed this PR

PR #1586, opened 2024-11-14
test on stable NC30

tobiasKaminsky:


tobiasKaminsky:
invalid share on NC30 fails due to wrong error handling on server side.
--> merging.

PR #1597, opened 2024-11-27
Inject Session Time Out To The Remote Operation Constructors

alperozturk96:
The client must be able to modify the session timeout according to their needs. For example, if the server instance slow, some network operations may fail. This could lead to issues, especially during end-to-end (E2E) operations.

**Changes**

SessionTimeOut data class has been added.
The default SessionTimeOut is utilized for each remote operation in the first constructor.
A second constructor has been added to allow custom SessionTimeOut injection for each remote operation.

**How to Test?**

Files client must build without any compile error
Files client must be able to configure SessionTimeOut for any remote operations

### DIFF
--- a/library/src/main/java/com/owncloud/android/lib/resources/e2ee/StoreMetadataV2RemoteOperation.kt
+++ b/library/src/main/java/com/owncloud/android/lib/resources/e2ee/StoreMetadataV2RemoteOperation.kt
@@ -26,7 +26,7 @@ class StoreMetadataV2RemoteOperation @JvmOverloads constructor(
     private val encryptedMetadataJson: String,
     private val token: String,
     private val signature: String,
-    var sessionTimeOut: SessionTimeOut = defaultSessionTimeOut
+    private val sessionTimeOut: SessionTimeOut = defaultSessionTimeOut
 ) : RemoteOperation<String>() {
 
     /**
--- a/library/src/main/java/com/owncloud/android/lib/resources/e2ee/StoreMetadataV2RemoteOperation.kt
+++ b/library/src/main/java/com/owncloud/android/lib/resources/e2ee/StoreMetadataV2RemoteOperation.kt
@@ -21,23 +21,13 @@ import org.json.JSONObject
 /**
  * Remote operation to store the folder metadata
  */
-class StoreMetadataV2RemoteOperation(
+class StoreMetadataV2RemoteOperation @JvmOverloads constructor(
     private val remoteId: String,
     private val encryptedMetadataJson: String,
     private val token: String,
-    private val signature: String
+    private val signature: String,
+    var sessionTimeOut: SessionTimeOut = defaultSessionTimeOut
 ) : RemoteOperation<String>() {
-    private var sessionTimeOut: SessionTimeOut = defaultSessionTimeOut
-
-    constructor(
-        remoteId: String,
-        encryptedMetadataJson: String,
-        token: String,
-        signature: String,
-        sessionTimeOut: SessionTimeOut
-    ) : this(remoteId, encryptedMetadataJson, token, signature) {
-        this.sessionTimeOut = sessionTimeOut
-    }
 
     /**
      * @param client Client object
--- a/library/src/main/java/com/owncloud/android/lib/resources/e2ee/StoreMetadataV2RemoteOperation.kt
+++ b/library/src/main/java/com/owncloud/android/lib/resources/e2ee/StoreMetadataV2RemoteOperation.kt
@@ -21,73 +21,76 @@ import org.json.JSONObject
 /**
  * Remote operation to store the folder metadata
  */
-class StoreMetadataV2RemoteOperation @JvmOverloads constructor(
-    private val remoteId: String,
-    private val encryptedMetadataJson: String,
-    private val token: String,
-    private val signature: String,
-    private val sessionTimeOut: SessionTimeOut = defaultSessionTimeOut
-) : RemoteOperation<String>() {
+class StoreMetadataV2RemoteOperation
+    @JvmOverloads
+    constructor(
+        private val remoteId: String,
+        private val encryptedMetadataJson: String,
+        private val token: String,
+        private val signature: String,
+        private val sessionTimeOut: SessionTimeOut = defaultSessionTimeOut
+    ) : RemoteOperation<String>() {
+        /**
+         * @param client Client object
+         */
+        @Deprecated("Deprecated in Java")
+        @Suppress("Detekt.TooGenericExceptionCaught")
+        override fun run(client: OwnCloudClient): RemoteOperationResult<String> {
+            var postMethod: Utf8PostMethod? = null
+            var result: RemoteOperationResult<String>
+            try {
+                // remote request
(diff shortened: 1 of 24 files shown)


AndyScherzinger (approved):

alperozturk96:
Failing test is unrelated. Example merged PR: https://github.com/nextcloud/android-library/pull/1593

library/src/main/java/com/owncloud/android/lib/resources/e2ee/StoreMetadataV2RemoteOperation.kt
```diff
@@ -25,6 +27,18 @@ class StoreMetadataV2RemoteOperation(
     private val token: String,
     private val signature: String
 ) : RemoteOperation<String>() {
+    private var sessionTimeOut: SessionTimeOut = defaultSessionTimeOut
+
+    constructor(
```
ZetaTom, line 32: Can't we use default constructor values in this case? That way we could consolidate both constructors into one:

```kotlin
class StoreMetadataV2RemoteOperation @JvmOverloads constructor(
    private val remoteId: String,
    private val encryptedMetadataJson: String,
    private val token: String,
    private val signature: String,
    private val sessionTimeOut: SessionTimeOut = defaultSessionTimeOut
) : RemoteOperation<String>() {
    // ...
}
```

We should be able to use default values in most Kotlin classes for this pull request.

library/src/main/java/com/owncloud/android/lib/resources/files/CreateFolderRemoteOperation.java
```diff
@@ -46,13 +46,27 @@ public class CreateFolderRemoteOperation extends RemoteOperation<String> {
...
+        this.sessionTimeOut = new SessionTimeOut(30000, 5000);
     }
 
     public CreateFolderRemoteOperation(String remotePath, boolean createFullPath, String token) {
         this(remotePath, createFullPath);
         this.token = token;
     }
 
+    public CreateFolderRemoteOperation(String remotePath, boolean createFullPath, SessionTimeOut sessionTimeOut) {
+        this.remotePath = remotePath;
```
ZetaTom, line 58: Is there a particular reason why we aren't just calling this constructor from the other ones with the default session timeouts?

library/src/main/java/com/owncloud/android/lib/resources/files/RemoveFileRemoteOperation.java
```diff
@@ -35,6 +35,12 @@ public class RemoveFileRemoteOperation extends RemoteOperation {
      */
     public RemoveFileRemoteOperation(String remotePath) {
         mRemotePath = remotePath;
+        sessionTimeOut = SessionTimeOutKt.getDefaultSessionTimeOut();
```
ZetaTom, line 38:
```suggestion
        sessionTimeOut = new SessionTimeOut(30000, 5000);
```

Regarding all of the non-default `30000` values: have we considered changing all of them to `40000`? Would this break anything?

Otherwise I would recommend leaving the timeouts as they are.

library/src/main/java/com/owncloud/android/lib/resources/files/RestoreFileVersionRemoteOperation.java
```diff
@@ -42,6 +43,13 @@ public class RestoreFileVersionRemoteOperation extends RemoteOperation {
     public RestoreFileVersionRemoteOperation(long fileId, String fileName) {
         this.fileId = fileId;
         this.fileName = fileName;
+        sessionTimeOut = SessionTimeOutKt.getDefaultSessionTimeOut();
```
ZetaTom, line 46:
```suggestion
        sessionTimeOut = new SessionTimeOut(30000, 5000);
```

library/src/main/java/com/owncloud/android/lib/resources/trashbin/RemoveTrashbinFileRemoteOperation.java
```diff
@@ -34,6 +34,12 @@ public class RemoveTrashbinFileRemoteOperation extends RemoteOperation {
      */
     public RemoveTrashbinFileRemoteOperation(String remotePath) {
         this.remotePath = remotePath;
+        sessionTimeOut = SessionTimeOutKt.getDefaultSessionTimeOut();
```
ZetaTom, line 37:
```suggestion
        sessionTimeOut = new SessionTimeOut(30000, 5000);
```

library/src/main/java/com/owncloud/android/lib/resources/trashbin/RestoreTrashbinFileRemoteOperation.java
```diff
@@ -42,6 +43,13 @@ public class RestoreTrashbinFileRemoteOperation extends RemoteOperation {
     public RestoreTrashbinFileRemoteOperation(String sourcePath, String fileName) {
         this.sourcePath = sourcePath;
         this.fileName = fileName;
+        sessionTimeOut = SessionTimeOutKt.getDefaultSessionTimeOut();
```
ZetaTom, line 46:
```suggestion
        sessionTimeOut = new SessionTimeOut(30000, 5000);
```

ZetaTom (commented):
Overall, this seems to be working as expected. There are a few minor things that I would suggest changing, for which I've added some comments.

library/src/main/java/com/owncloud/android/lib/resources/e2ee/StoreMetadataV2RemoteOperation.kt
alperozturk96: These classes called from java files, and you have to pass sessionTimeOut parameter, thus default value cannot be used from java files. Client code shouldn't affect from these changes.

ZetaTom: While I fully agree, the [`@JvmOverloads`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-overloads/) annotation used in the example instructs the compiler to generate all necessary constructors, including ones without default values.

alperozturk96: https://github.com/nextcloud/android-library/pull/1597/commits/635e112725fe1f195039891abcedd33fb7d2a8e2 Thanks

tobiasKaminsky (approved):

alperozturk96:
@tobiasKaminsky @AndyScherzinger Can we merge this PR? Since unrelated tests are keep failing for previously merged PRs as well.

ZetaTom (commented):
Apart from the new default timeouts, everything looks fine. I haven't noticed any change in behaviour when running tests. However, we need to consider whether the new timeouts have any unintended consequences.

alperozturk96:
Increasing SYNC_READ_TIMEOUT and SYNC_CONNECTION_TIMEOUT values should not cause any issue. 

If we have any solid argument against it, please share. Any code changes can cause unintended consequences.

PR #1631, opened 2025-01-15
Support Files Download Limits

ZetaTom:
This pull request implements all operations required to interface with the [files_downloadlimit](https://github.com/nextcloud/files_downloadlimit) server app which allows users to set limit of how many times a file may be downloaded from a public share.

The UI implementation will follow in a separate pull request.

See nextcloud/files-clients#54 for more information.


library/src/main/java/com/owncloud/android/lib/resources/status/OwnCloudVersion.java
```diff
@@ -24,6 +24,7 @@ public class OwnCloudVersion implements Comparable<OwnCloudVersion>, Parcelable
     public static final OwnCloudVersion nextcloud_18 = new OwnCloudVersion(0x12000000); // 18.0
     public static final OwnCloudVersion nextcloud_19 = new OwnCloudVersion(0x13000000); // 19.0
     public static final OwnCloudVersion nextcloud_20 = new OwnCloudVersion(0x14000000); // 20.0
+    public static final OwnCloudVersion nextcloud_30 = new OwnCloudVersion(0x1E000000); // 30.0
```
tobiasKaminsky, line 27: We have NextcloudVersion for that

tobiasKaminsky (changes requested):
Use NextcloudVersion

tobiasKaminsky (approved):

PR #1645, opened 2025-01-22
Bump minimum required server version

ZetaTom:
Bump minimum server version in accordance with Nextcloud files client.


library/src/main/java/com/owncloud/android/lib/resources/status/OwnCloudVersion.java
```diff
@@ -115,14 +114,6 @@ static protected int getParsedVersion(String version) throws NumberFormatExcepti
         return versionValue;
     }
 
-    public boolean isShareesOnDavSupported() {
```
tobiasKaminsky, line 118: This needs a corresponding PR on files app

tobiasKaminsky (approved):

PR #1683, opened 2025-03-06
Use SHA-256

alperozturk96:
### Fresh Install

https://github.com/user-attachments/assets/01376eca-e036-4fd9-900a-d7affe91e2ff

### Upgrade from 3.30.8

https://github.com/user-attachments/assets/10f28582-f920-4171-aa7a-8e88ad6dda8f


tobiasKaminsky (changes requested):
please test 
- create CSR with Android
- private/public key is generated and stored on server
- Desktop and iOS needs to understand it / be able to add it as new account

alperozturk96:
All tests between clients have been completed.

PR #1706, opened 2025-04-10
Add Attributes to Share

alperozturk96:


tobiasKaminsky:
REUSE fails, as license is not MIT.

alperozturk96:
@tobiasKaminsky We've noticed that these tests are now consistently failing across all other PRs as well. Could it be that the test account configuration on the test server has changed?

Many of our Android Library tests rely on scenarios like this, where an admin attempts to create a share with another admin. This is expected to fail with error code SHARE_FORBIDDEN.

However, when I modify the relevant code locally so that the admin creates a share with a "user" instead, the tests pass successfully.

```
CreateShareRemoteOperation(
                "/note/",
                ShareType.USER,
                "admin",
                false,
                "",
                OCShare.MAXIMUM_PERMISSIONS_FOR_FOLDER,
                true,
                "",
                null
            ).execute(client)

```

Screenshot 2025-05-06 at 11 36 58

tobiasKaminsky:
The share errors are fixed and merged.
Can you rebase?

alperozturk96:
@tobiasKaminsky Done, thank you for the fix.

library/src/androidTest/java/com/owncloud/android/lib/resources/shares/CreateShareRemoteOperationIT.kt
```diff
@@ -22,40 +24,60 @@ class CreateShareRemoteOperationIT : AbstractIT() {
...
+        val attributes =
+            ShareAttributesJsonHandler.toJson(
+                listOf(ShareAttributes.createDownloadAttributes(true))
+            )
+        val note = "Note with attributes"
+        val path = "/shareWithAttributes/"
+
+        createFolder(path)
+        val share = createShare(path, "admin", note, attributes)
+        assertEquals(note, share.note)
```
ZetaTom, line 44: I think that this test case should also check if the share attributes have been set correctly. Maybe something like this could work:

```suggestion
        val attributes = listOf(ShareAttributes.createDownloadAttributes(true))
        val note = "Note with attributes"
        val path = "/shareWithAttributes/"

        createFolder(path)
        val share = createShare(path, "admin", note, ShareAttributesJsonHandler.toJson(attributes))
        assertEquals(note, share.note)
        assertEquals(attributes, ShareAttributesJsonHandler.toList(share.attributes))
```

ZetaTom (commented):
In general everything seems to work. One minor thing is that we're not actually checking that the attributes have been set correctly in any test case.

PR #1721, opened 2025-05-08
Fix - AbstractIT

alperozturk96:
Resolved issues encountered during local environment testing with AbstractIT. 

1. The test was forcing the use http instead of https, which caused the waitForServer() function to fail. 

2. The removeOnClient() method was unable to delete files as expected, leading to test failures. File can be locked, thus remove will fail.

### DIFF
--- a/library/src/androidTest/java/com/owncloud/android/AbstractIT.java
+++ b/library/src/androidTest/java/com/owncloud/android/AbstractIT.java
@@ -10,7 +10,6 @@ package com.owncloud.android;
 
 import static junit.framework.TestCase.assertTrue;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
 import static org.junit.Assume.assumeTrue;
 
 import android.content.Context;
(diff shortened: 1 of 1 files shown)


alperozturk96:
ToggleEncryptionRemoteOperation fails locally even with these changes.

I'm getting 403, FORBIDDEN.

Local instance username is "user"

ZetaTom:
> ToggleEncryptionRemoteOperation in removeOnClient() fails locally, even with these changes, but works on our server.
> 
> I'm getting a 403 Forbidden error.
> 
> The username for the local instance is "user".

During testing we've noticed that this does indeed happen while running test cases explicitly. I've also been able to replicate this on my machine. However, when placing a breakpoint in `ToggleEncryptionRemoteOperation` and manually unlocking a folder in the app, a 200 status code is returned instead.

library/src/androidTest/java/com/owncloud/android/AbstractIT.java
```diff
@@ -265,30 +267,57 @@ public void after() {
...
 
-                assertTrue("Failed to remove " + remoteFile.getRemotePath(),
-                    new RemoveFileRemoteOperation(remoteFile.getRemotePath()).execute(client).isSuccess());
+            if (remoteFile.isLocked() && remotePath != null) {
+                unlockRemoteFile(remotePath);
             }
+
+            boolean isRemoteFileRemoved = removeRemoteFile(remotePath);
+            final var removeFileOperationErrorMessage = ("Failed to remove " + remotePath);
+            assertTrue(removeFileOperationErrorMessage, isRemoteFileRemoved);
```
ZetaTom, line 296:
```suggestion
            assertTrue("Failed to remove " + remotePath, isRemoteFileRemoved);
```

AndyScherzinger (approved):

PR #1724, opened 2025-05-10
Fixing a data transfer error

Alexander-Ger-Reich:
The -1 can cause the last byte to not end up in the request body (incomplete file), causing the server to time out because it expects the body length to be one byte larger based on the content length. However, this one byte that isn't sent is present in the file.

I was able to successfully test this and it fixed the error for me.

Fix for
image


Alexander-Ger-Reich:
Fix for
image

Alexander-Ger-Reich:
length starts at 0 and not at 1, so you should not use -1 here

joshtrichards:
May fix:

- nextcloud/android#14950
- https://github.com/nextcloud/android/issues/12990#issuecomment-2252882308

Also worth mentioning owncloud/android-library#436. While not directly related, something we should confirm while we're in this area of code since it's a similar matter.

tobiasKaminsky (approved):

Alexander-Ger-Reich:
private ByteBuffer mBuffer = ByteBuffer.allocate(4096);

ZetaTom:
> Noob question, but how long does it take for changes in nextcloud:master to make it to the android client in Google Play?

We've created a backport of this change to `stable-2.21` which was merged yesterday. Version 2.21 will be used for the [3.32.0 release](https://github.com/nextcloud/android/milestone/107) of the android files client. Currently, we're still working out some minor issues but plan to release this version soon.

PR #1741, opened 2025-06-11
BugFix - Do not Filter Share Type

alperozturk96:
Client PR: https://github.com/nextcloud/android/pull/14989

### DIFF
--- a/library/src/main/java/com/owncloud/android/lib/common/network/WebdavEntry.kt
+++ b/library/src/main/java/com/owncloud/android/lib/common/network/WebdavEntry.kt
@@ -73,7 +73,7 @@ class WebdavEntry constructor(
     var unreadCommentsCount = 0
     var isHasPreview = false
     var note = ""
-    var sharees = arrayOfNulls<ShareeUser>(0)
+    var sharees = arrayOf<ShareeUser>()
     var richWorkspace: String? = null
     var isLocked = false
         private set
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/owncloud/android/lib/common/network/WebdavEntry.kt
```diff
@@ -548,20 +544,11 @@ class WebdavEntry constructor(
...
-            ShareType.EMAIL == shareType ||
-                ShareType.FEDERATED == shareType ||
-                ShareType.GROUP == shareType ||
-                ShareType.ROOM == shareType
-        return if ((isSupportedShareType || displayName.isNotEmpty()) && userId.isNotEmpty()) {
-            ShareeUser(userId, displayName, shareType)
-        } else {
-            null
-        }
+        return ShareeUser(userId, displayName, shareType)
```
ZetaTom, line 551: This changes what this function returns in some cases. For example, if a public link share was parsed, `null` was previously returned, now `ShareeUser("", "", PUBLIC_LINK)` is returned. I'm not sure if or what implications this change has.

alperozturk96: This change is intentional. Previously, the corresponding remote file lacked a shareType due to a missing username. Now, even if the username is absent, we can still determine the shareType, allowing it to be displayed in the UI.

A missing username should not cause issues, as external share links do not include one, and the username should not be mandatory. We might consider renaming the data class and adding documentation in a separate PR.

See, there is no username for this share.

Screenshot 2025-06-13 at 09 03 57

PR #1745, opened 2025-06-16
Fix data transfer error

ZetaTom:
> The -1 can cause the last byte to not end up in the request body (incomplete file), causing the server to time out because it expects the content length to be one byte larger. However, this one byte that isn't loaded and isn't sent is present in the file.
> 
> I was able to successfully test this and it fixed the error for me.

This pull request is solely intended to run all CI tests of #1724, which is a community contribution by @Alexander-Ger-Reich. Due to our security policies, we cannot run CI tests on forked repositories. Consequently, all changes have been pushed to a branch within this repository. This pull request shall be closed in favour of #1724 once all tests have been completed.

### DIFF
--- a/library/src/main/java/com/owncloud/android/lib/common/network/ChunkFromFileChannelRequestEntity.java
+++ b/library/src/main/java/com/owncloud/android/lib/common/network/ChunkFromFileChannelRequestEntity.java
@@ -103,7 +103,7 @@ public class ChunkFromFileChannelRequestEntity implements RequestEntity, Progres
             if (size == 0) {
                 size = -1;
             }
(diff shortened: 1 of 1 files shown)


crimsonflame123:
Did this one also hit a roadblock?

Alexander-Ger-Reich:
> Did this one also hit a roadblock?

What does Analysis do and why does it fail?

crimsonflame123:
> > Did this one also hit a roadblock?
> 
> What does Analysis do and why does it fail?

I'm new to this codebase but have a look at this. Warning count has increased and theres also message about stuff becoming an error in Kotlin.

https://github.com/nextcloud/android-library/actions/runs/15683089591/job/44179251411?pr=1745

PR #1761, opened 2025-07-01
BugFix - Filter Only One TextInput and one TextOutput Task Types

alperozturk96:
Currently, on the Android client, we are retrieving task types from the `/ocs/v2.php/taskprocessing/tasktypes` endpoint, and it only uses the "text" task types. However, the Chat API task type is also served from this endpoint, which requires more than a single input string, as seen here:

https://github.com/nextcloud/server/blob/master/lib/public/TaskProcessing/TaskTypes/TextToTextChat.php#L70-L85.

Client & Library should filter **only one text input and one text output.**

Client PR: [BugFix - Filter Only One TextInput and one TextOutput Task Types #15079](https://github.com/nextcloud/android/pull/15079)

### How to test?

1. Set up two instances: one with Nextcloud 28 and another with Nextcloud version 30 or higher.
2. Enable assistant app from web client
3. Create a task and user should not be able to see the Chat task type.

### DIFF
--- a/library/src/main/java/com/owncloud/android/lib/resources/assistant/v2/model/TaskTypes.kt
+++ b/library/src/main/java/com/owncloud/android/lib/resources/assistant/v2/model/TaskTypes.kt
@@ -14,14 +14,14 @@ data class TaskTypes(
 
 data class TaskTypeData(
     val id: String?,
-    val name: String?,
+    val name: String,
     val description: String?,
-    val inputShape: Map<String, Shape>?,
-    val outputShape: Map<String, Shape>?
+    val inputShape: Map<String, Shape>,
+    val outputShape: Map<String, Shape>
 )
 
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/owncloud/android/lib/resources/assistant/v2/model/TaskTypes.kt
```diff
@@ -16,20 +16,12 @@ data class TaskTypeData(
...
-    val output: Shape?
+    val inputShape: Map<String, Shape>?,
+    val outputShape: Map<String, Shape>?
 )
 
 data class Shape(
-    val name: String,
-    val description: String,
-    val type: String
+    val name: String?,
```
tobiasKaminsky, line 24: double check nullable?

PR #1813, opened 2025-08-28
Drop NC19

tobiasKaminsky:


AndyScherzinger:
@tobiasKaminsky looking at the original if statements you should be keeping the if branch not the else branch, no?

tobiasKaminsky:
> @tobiasKaminsky looking at the original if statements you should be keeping the if branch not the else branch, no

you are right!

tobiasKaminsky:
Fixed

alperozturk96 (changes requested):
Screenshot 2025-09-03 at 14 08 51

alperozturk96 (approved):

PR #1836, opened 2025-09-22
Declarative UI

tobiasKaminsky:


library/src/main/java/com/nextcloud/android/lib/resources/clientintegration/App.kt
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+package com.nextcloud.android.lib.resources.clientintegration
+
+import com.google.gson.annotations.SerializedName
+
+data class App(
+    val version: Double,
+    @SerializedName("context-menu")
```
alperozturk96: Why this is not using `kotlinx.parcelize.Parcelize`  as well ?

library/src/main/java/com/nextcloud/android/lib/resources/clientintegration/Row.kt
```diff
@@ -0,0 +1,16 @@
+/*
+ * Nextcloud Android Library
+ *
+ * SPDX-FileCopyrightText: 2025 Your Name <your@email.com>
```
alperozturk96: Please update license header

alperozturk96 (changes requested):

library/src/main/java/com/nextcloud/android/lib/resources/clientintegration/GetClientIntegrationJsonOperation.kt
```diff
@@ -0,0 +1,81 @@
...
+                )
+            val status = client.execute(getMethod)
+            if (status == HttpStatus.SC_OK) {
+                val terms =
+                    getServerResponse(
+                        getMethod,
+                        object : TypeToken<ServerResponse<JsonArray>>() {}
+                    )?.ocs?.data
+
+                val json = parseResult(getMethod.getResponseBodyAsString())
```
alperozturk96: json result is not used

library/src/main/java/com/owncloud/android/lib/resources/status/Method.kt
```diff
@@ -0,0 +1,15 @@
...
+ * Nextcloud Android Library
+ *
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-FileCopyrightText: 2025 Tobias Kaminsky <tobias@kaminsky.me>
+ * SPDX-License-Identifier: MIT
+ */
+
+package com.owncloud.android.lib.resources.status
+
+enum class Method {
```
alperozturk96: This needs to be parcelable since parent class is also?

```
import android.os.Parcelable
import kotlinx.parcelize.Parcelize

@Parcelize
enum class Method : Parcelable {
    GET,
    POST,
    PUT
}

```

alperozturk96 (changes requested):

library/src/main/java/com/nextcloud/android/lib/resources/clientintegration/ClientIntegrationUI.kt
```diff
@@ -0,0 +1,18 @@
...
+
+package com.nextcloud.android.lib.resources.clientintegration
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+data class ClientIntegrationUI(
+    val version: Double,
+    val root: Layout
```
alperozturk96: During my tests layout was null and caused crash in files app.

```
 E  FATAL EXCEPTION: DefaultDispatcher-worker-2
                                                                                                    Process: com.nextcloud.client, PID: 4249
                                                                                                    java.lang.NullPointerException: Attempt to invoke virtual method 'void com.nextcloud.android.lib.resources.clientintegration.Layout.writeToParcel(android.os.Parcel, int)' on a null object reference
```

alperozturk96 (changes requested):

alperozturk96 (approved):

library/src/main/java/com/nextcloud/android/lib/resources/clientintegration/App.kt
tobiasKaminsky: There is no need for it. 
Serialiazed name is needed to parse json.

PR #1949, opened 2026-02-12
Add operations for web push

p1gp1g:
Now Notifications app support webpush, we can use it with the Android apps: https://github.com/nextcloud/notifications/pull/2662

This PR adds:
- Capability for web push
- Operations to get the server VAPID key, registration/activation/unregistration for web push notifications

### DIFF
--- a/library/src/main/java/com/owncloud/android/lib/resources/notifications/ActivateWebPushRegistrationOperation.kt
+++ b/library/src/main/java/com/owncloud/android/lib/resources/notifications/ActivateWebPushRegistrationOperation.kt
@@ -1,0 +1,63 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2026 Simon Gougeon <git@sgougeon.fr>
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+package com.owncloud.android.lib.resources.notifications
+
+import com.nextcloud.common.JSONRequestBody
+import com.nextcloud.common.NextcloudClient
+import com.nextcloud.operations.PostMethod
+import com.owncloud.android.lib.common.operations.RemoteOperation
+import com.owncloud.android.lib.common.operations.RemoteOperationResult
+import com.owncloud.android.lib.common.utils.Log_OC
+import org.apache.commons.httpclient.util.HttpURLConnection
+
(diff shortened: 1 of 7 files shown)


Elsensee (commented):
The code looks good to me. It's just an http client for the consumer.

Only a few nitpicks:
- I think, generally, there should be an empty newline at the end of a file.
- You should add yourself to the copyright in the header file of the newly added files. :)
- You have put `// OCS Route` just above the `const` with the same name. That is unnecessary.
- Why did you put the new classes under `com.owncloud.android.lib`? The owncloud namespace feels very legacy to me and not like there should be new code put there, but I am none of the code owners, so I can only wonder why you decided to not put it into the nextcloud namespace?

AndyScherzinger:
> The CI complain about some "TooGenericExceptionCaught", but I do exactly like ProxyOperations - I suppose it can be ignored.

Best to add a Suppression annotation 👍

library/src/main/java/com/owncloud/android/lib/resources/notifications/ActivateWebPushRegistrationOperation.kt
```diff
@@ -0,0 +1,63 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2025 Your Name <your@email.com>
```
alperozturk96: Please update license headers.

library/src/main/java/com/owncloud/android/lib/resources/notifications/models/VapidResponse.kt
```diff
@@ -0,0 +1,12 @@
+/*
```
alperozturk96, line 1: Can you rebase to latest master?

alperozturk96 (changes requested):

library/src/main/java/com/owncloud/android/lib/resources/notifications/models/VapidResponse.kt
p1gp1g: I've just rebased the branch

p1gp1g:
I've squashed the commits, updated the file headers, and rebased the PR

tobiasKaminsky:
@mahibi do you have similar code in Talk?
Does it make sense to centralize this a bit?

PR #2000, opened 2026-04-29
feat(e2ee): add deletion functions

alperozturk96:
App PR: https://github.com/nextcloud/android/pull/16926


library/src/androidTest/java/com/owncloud/android/lib/resources/users/DeleteE2ERemoteOperationIT.kt
```diff
@@ -0,0 +1,36 @@
...
+package com.owncloud.android.lib.resources.users
+
+import com.owncloud.android.AbstractIT
+import com.owncloud.android.lib.resources.e2ee.DeleteEncryptedFilesRemoteOperation
+import junit.framework.TestCase.assertTrue
+import org.junit.Test
+
+class DeleteE2ERemoteOperationIT : AbstractIT() {
+    @Test
+    fun testDeleteEncryptedFiles() {
```
tobiasKaminsky, line 17: Those tests will also return success if there is no key/file at all, right?
I guess it is hard to test it within library, but then we should test it within client:
- set up everything
- doing existing e2e tests
- remove files and keys
- verify that they are not there, via get private/public key

tobiasKaminsky (approved):

PR #2047, opened 2026-06-16
chore(sdk): min version 26

alperozturk96:
### Reason

Use modern APIs to provide faster, safer, and more reliable networking for clients, while reducing maintenance overhead and avoiding duplicate code.

### Example

Using a `ConcurrentHashMap` instead of a `synchronized` block with `HashSet`.

### Changes

Min supported Android version is now 8.0.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -143,7 +143,7 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 23
+        minSdkVersion 26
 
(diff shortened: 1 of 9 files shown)


alperozturk96:
image

@tobiasKaminsky ?

AndyScherzinger (dismissed):

AndyScherzinger (approved):

PR #2056, opened 2026-06-19
chore(build): update

alperozturk96:
### Changes

- Uses built-in kotlin
- Removes deprecated gradle properties
- Reduces ram usage
- Enabled parallel sync for Gradle 9.4+

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -159,7 +159,7 @@ android {
 
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_21
-        targetCompatibility JavaVersion.VERSION_21
+        targetCompatibility JavaVersion.VERSION_17
     }
 
     kotlin {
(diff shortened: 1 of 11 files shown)


alperozturk96:
only fails due to code ql kotlin version:

Screenshot 2026-07-17 at 11 16 43

library/build.gradle
```diff
@@ -159,12 +158,14 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility JavaVersion.VERSION_21
+        targetCompatibility JavaVersion.VERSION_21
```
tobiasKaminsky: Can this be a problem for any app using our code? 
If they are on 17?

alperozturk96: sourceCompatibility = what language we write
targetCompatibility = what language level we produce -> this could break the clients 

when i only set the targetCompatibility to 17 ci fails. thus better to keep 17.


Screenshot 2026-07-21 at 13 35 36

tobiasKaminsky (approved):

PR #2102, opened 2026-08-03
improve memory footprint for read operations

alperozturk96:
### Changes

- Catches OOME
- Read only once `remoteData.getResponses();`


daniele-verducci:
Hey @alperozturk96 , I was able to replicate the OOM crash using the current prod library on a directory with 12k files, and can confirm that **with the library from this PR the app does NOT crash**.
However, I get no feedback about the issue on the phone's screen, I just see "Loading..." and after a while "No files here".

image

I can see in the ADB logs the correct error:
```
Synchronized /InstantUpload/OpenCamera/: Unexpected exception
                                                                                                    java.lang.Exception: Not enough memory to read the contents of /InstantUpload/OpenCamera/
```
but nothing on the screen, so I fear that to an unbeknowing user, it may seem that the directory is effectively empty (or that the content has gone lost).

alperozturk96:
> Hey @alperozturk96 , I was able to replicate the OOM crash using the current prod library on a directory with 12k files, and can confirm that **with the library from this PR the app does NOT crash**. However, I get no feedback about the issue on the phone's screen, I just see "Loading..." and after a while "No files here".
> 
> image
> I can see in the ADB logs the correct error:
> 
> ```
> Synchronized /InstantUpload/OpenCamera/: Unexpected exception
>                                                                                                     java.lang.Exception: Not enough memory to read the contents of /InstantUpload/OpenCamera/
> ```
> 
> but nothing on the screen, so I fear that to an unbeknowing user, it may seem that the directory is effectively empty (or that the content has gone lost).

This needs to be handled in client side. I will expose OOME then handle in client via different PR.

daniele-verducci:
Right, sorry 🤦

daniele-verducci (approved):

alperozturk96:
> Right, sorry 🤦

Client PR: https://github.com/nextcloud/android/pull/17479

PR #2130, opened 2026-09-01
improve chunked-upload timeout

alperozturk96:
Tested with 2.5mb, 121.5mb, 245mb, 2GB

### Changes

- Adjust the assemble time
- Uses `Kotlin.Duration` instead `Int`.
- Add tests for different filesizes and assembled time

### DIFF
--- a/library/src/test/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperationTest.kt
+++ b/library/src/test/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperationTest.kt
@@ -8,6 +8,9 @@
  */
 package com.owncloud.android.lib.resources.files
 
+import com.owncloud.android.lib.resources.files.ChunkedFileUploadRemoteOperation.Companion.ASSEMBLE_TIME_BASE
+import com.owncloud.android.lib.resources.files.ChunkedFileUploadRemoteOperation.Companion.ASSEMBLE_TIME_MAX
+import com.owncloud.android.lib.resources.files.ChunkedFileUploadRemoteOperation.Companion.ASSEMBLE_TIME_PER_GB
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotEquals
 import org.junit.Test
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
Bette to have it in, so that no server times out for us ✔

library/src/test/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperationTest.kt
```diff
@@ -24,51 +27,74 @@ class ChunkedFileUploadRemoteOperationTest {
...
         // 100b
         Mockito.`when`(file.length()).thenReturn(100L)
-        assertEquals(sut.assembleTimeMin, sut.calculateAssembleTimeout(file))
+        assertEquals(ASSEMBLE_TIME_BASE, ChunkedFileUploadRemoteOperation.calculateAssembleTimeout(file))
 
         // 1Mb
         Mockito.`when`(file.length()).thenReturn(1 * MB)
-        assertEquals(sut.assembleTimeMin, sut.calculateAssembleTimeout(file))
+        assertEquals(
+            ASSEMBLE_TIME_BASE + ASSEMBLE_TIME_PER_GB / 1000,
```
tobiasKaminsky, line 42: expected value should be hardcoded, else it does the same as calc function and thus if there is problem, it will not be seen?

tobiasKaminsky:
> * Increases the min assemble time to 3 min

I only see 1min as min?

library/src/test/java/com/owncloud/android/lib/resources/files/ChunkedFileUploadRemoteOperationTest.kt
alperozturk96: Already like that. I don't get what you mean.

Image

tobiasKaminsky: x + y / 1000 is not hardcoded.
I would rather use 3, so that it does not depend on any calculation.

In your initial description you wrote: 
min assemble time to 3 min

With "your" test it was wrong/not clear that there is a diff between code and desc.

So having real values, from 
- 3 minimum
- 13 1Gb
- 26 2Gb
- 60 maximum

makes it easier to read, and also reveals any logic error, if computation of it changes?

alperozturk96: I got what you mean but having magic numbers in tests makes the test more unclear to me.

These are ASSEMBLE_TIME_BASE, ASSEMBLE_TIME_PER_GB constants. If we change those also magic numbers in the tests needs to be changed.

Those constants part of the production logic. Purpose of the assert is here are we getting expected duration against the calculated assemble time out.

e.g. for 1GB file -> should be ASSEMBLE_TIME_BASE + ASSEMBLE_TIME_PER_GB and calculateAssembleTimeout must return same value.

ASSEMBLE_TIME_BASE is 1.minutes
ASSEMBLE_TIME_PER_GB is 10.minutes
ASSEMBLE_TIME_MAX is 1.hour

so 11.minutes * 1 = 11 and min of 11.minutes and 1.hour is 11.minutes  why we should write -> 11 * 60 * 1000?

tobiasKaminsky (approved):

daniele-verducci (approved):

